PR #6, opened 2021-06-16
[MRG] Adding connectivity data structure implemented with xarray backbone

adam2392:
Closes: #13 

Begin development of a connectivity data structure that holds connectivity data.

Also updated the circleCI for attempts at building docs for `dev/` and building docs for `stable/`.

Also removed GH actions for style checking for all python versions and now doing only for Python3.6.

TODO:

- [x] Refactor existing methods to use this
- [ ] ~~Add an example to show-case it~~
- [x] Adjust the unit tests to work with the api change
- [x] Adjust the docs to work with the api change


adam2392:
My opinion is perhaps I will probably keep the connectivity containers rather lightweight for now. They are literally a structured lumpy array for all intents and purposes (which makes me think... should we be using xarray?).

We can tack on the plotting and saving next, which will greatly simplify user life. Then simultaneously add the io to load in a connectivity dataset, so its full loop.

doc/whats_new.rst
```diff
@@ -22,15 +22,24 @@ Current
...
 
+- Adding :class:`mne_connectivity.TemporalConnectivity`, :class:`mne_connectivity.SpectralConnectivity` and :class:`mne_connectivity.SpectroTemporalConnectivity` as a data structure to hold connectivity data, by `Adam Li`_ (:gh:`6`)
+
 Bug
 ~~~
 
 API
 ~~~
 
+- :func:`mne_connectivity.envelope_correlation` has a new required parameter, ``names``, which specifies the name of the signals passed in, by `Adam Li`_ (:gh:`6`)
```
larsoner, line 33: I would not require this. We can just have `names=None` behave like `map(range(n_signals), str)`

mne_connectivity/base.py
```diff
@@ -0,0 +1,172 @@
...
+        self.times = times
+        self.n_epochs = n_epochs
+
+    @property
+    def frequency_axis(self):
+        return 2
+
+    @property
+    def time_axis(self):
+        return 3
```
larsoner, line 142: If we can standardize on time axis being `-1` and freq axis being `-2` then we don't need these.

Also if we really want axis labels then we could just have `def axis_labels(self): return ('epochs', 'frequencies', 'time')` or so. This is starting to feel a little bit like XArray territory...

@agramfort you've opposed using XArray in the past, WDYT about trying it out in mne-connectivity? We could have `object.data` return an `ndarray` cast of the internally stored xarray, and have `object.xarray` for example give the root xarray object.

adam2392: However, sometimes the frequency axis is not there, and sometimes the time-axis is not there?

larsoner: Having vs not having properties is a bit ugly because it leads to code like `isinstance(...)` or `hasattr(...)`, might be nicer just to have a `.dimensions` property that is a tuple of str. Then if you want the axis number you can always do `obj.dimensions.index('time')` for example. But I would look at the xarray API for inspiration/naming here

adam2392: Makes sense! For my reference: http://xarray.pydata.org/en/stable/user-guide/terminology.html

looks like: `conn.dims` will show `('node', 'node', 'time', ...)`

larsoner: Sounds good, though we should probably call it `node_from` and `node_to` to indicate potential directionality

adam2392: Ooo yeah that sounds way better!

adam2392:
@larsoner can you take a look at the CircleCI changes here and lmk if that's what you had in mind?

The only question is once there is a new maintenance build (e.g. `maint/0.2`), then how do I take the existing `stable/` docs folder in `gh-pages` and rename them to `maint/0.1`? Do I do that manually?

drammock:
+1 for trying out `xarray` here.  It's something I've often wished for in MNE-Python and I would be much more willing to take the plunge if some of the pitfalls could be worked out in advance on a smaller subset of our codebase.  This seems like a good test case if @adam2392 is up for it.  That said, if it looks like too big of a rabbit hole, I wouldn't want it to derail getting the actual migration done.

larsoner:
Yes during release of 0.2 you copy `stable/` to `0.1/`, you should see a step like this in the MNE-Python release wiki instructions

larsoner:
And yes that's about what I'd expect for CircleCI

adam2392:
> +1 for trying out `xarray` here. It's something I've often wished for in MNE-Python and I would be much more willing to take the plunge if some of the pitfalls could be worked out in advance on a smaller subset of our codebase. This seems like a good test case if @adam2392 is up for it. That said, if it looks like too big of a rabbit hole, I wouldn't want it to derail getting the actual migration done.

I'll file this as a new issue then to see if its possible. I'm not as familiar with Xarray, so I will try keeping the containers pretty lightweight for now.

mne_connectivity/base.py
agramfort: no objection for xarray here. I think it's a good fit for the usecase here.

doc/whats_new.rst
britta-wstnr: Yes, especially for source points this is what you would  end up doing anyway

mne_connectivity/base.py
```diff
@@ -0,0 +1,211 @@
...
+
+from mne.utils import sizeof_fmt, object_size
+
+
+class _Connectivity():
+    def __init__(self, data, names, method, indices):
+        """Base class container for connectivity data.
+
+        Connectivity data is anything represents "connections"
+        between nodes as a (N, N) array. It can be symmetric, or
```
britta-wstnr, line 11:
```suggestion
        Connectivity data is anything that represents "connections"
        between nodes as a (N, N) array. It can be symmetric, or
```

mne_connectivity/base.py
```diff
@@ -0,0 +1,211 @@
...
+        asymmetric (if it is symmetric, storage optimization will
+        occur). In addition,
```
britta-wstnr, line 13: sentence missing?

mne_connectivity/tests/test_spectral.py
```diff
@@ -91,14 +91,21 @@ def test_spectral_connectivity(method, mode):
...
+        if isinstance(method, list):
+            freqs = con[0].freqs
+            n = con[0].n_epochs
+        else:
+            freqs = con.freqs
+            n = con.n_epochs
+
         assert (n == n_epochs)
-        assert_array_almost_equal(times_data, times)
+        # assert_array_almost_equal(times_data, times)
```
britta-wstnr, line 108: I think it's probably still a good idea to test that the time vector doesn't get mangled in the connectivity function.

adam2392: So it looks like the `times_data` vector is just the tmin:tmax inside `_prepare_connectivity()`:

```
    mask = _time_mask(times_in, tmin, tmax, sfreq=sfreq)
    tmin_idx, tmax_idx = np.where(mask)[0][[0, -1]]
    tmax_idx += 1
    tmin_true = times_in[tmin_idx]
    tmax_true = times_in[tmax_idx - 1]  # time of last point used

    times = times_in[tmin_idx:tmax_idx]
```

Is it necessary to keep that?

adam2392:
I've added some changes to incorporate xarray-like api now.

Any thoughts @larsoner @britta-wstnr ?

adam2392:
Okay now everything works internally, imo is backwards compatible but now the sphinx documentation building now breaks:

```
sphinx-build -b html -d _build/doctrees  -nWT --keep-going . _build/html
Running Sphinx v4.0.2
making output directory... done
checking for /home/circleci/project/doc/references.bib in bibtex cache... not found
parsing bibtex file /home/circleci/project/doc/references.bib... parsed 9 entries
[autosummary] generating autosummary for: api.rst, bibliography.rst, index.rst, install.rst, whats_new.rst
[autosummary] generating autosummary for: /home/circleci/project/doc/generated/mne_connectivity.envelope_correlation.rst, /home/circleci/project/doc/generated/mne_connectivity.phase_slope_index.rst, /home/circleci/project/doc/generated/mne_connectivity.spectral_connectivity.rst
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://mne.tools/dev/objects.inv...
loading intersphinx inventory from https://mne.tools/mne-bids/dev/objects.inv...
loading intersphinx inventory from https://numpy.org/devdocs/objects.inv...
loading intersphinx inventory from https://scipy.github.io/devdocs/objects.inv...
loading intersphinx inventory from https://matplotlib.org/objects.inv...
loading intersphinx inventory from https://pandas.pydata.org/pandas-docs/dev/objects.inv...
loading intersphinx inventory from https://scikit-learn.org/stable/objects.inv...
loading intersphinx inventory from https://docs.pyvista.org/objects.inv...
loading intersphinx inventory from https://joblib.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://nipy.org/nibabel/objects.inv...
loading intersphinx inventory from http://nilearn.github.io/objects.inv...
generating gallery...
Using Sphinx-Gallery to convert rst text blocks to markdown for .ipynb files.
generating gallery for auto_examples... [ 11%] mne_inverse_psi_visual.py
make: *** [Makefile:56: html] Killed
```

larsoner:
CircleCI kills things when memory usage is too high. To me it suggests that some of the memory optimizations / list / generator support from MNE-Python's code and example were removed

larsoner:
... and locally if you run the example with `mprof run ...; mprof plot` or just look at a process manager while it builds, you can see if this is the case or not

mne_connectivity/spectral.py
```diff
@@ -916,19 +926,63 @@ def spectral_connectivity(data, method='coh', indices=None, sfreq=2 * np.pi,
...
+        # keep all arrays as all-to-all connectivity with nans
+        logger.info('    assembling connectivity matrix for subset')
+        con_flat = con
+        con = list()
+        for this_con_flat in con_flat:
+            this_con = np.zeros((n_signals, n_signals) +
+                                this_con_flat.shape[1:],
+                                dtype=this_con_flat.dtype)
+            this_con[:] = np.nan
+            this_con[indices_use] = this_con_flat
```
adam2392, line 941: @larsoner do you think this could be messing w/ the RAM since it might scale up the amount of RAM needed when the user only selects a few indices?

larsoner: Yeah it could. So maybe we do need to efficiently store the connectivity. Maybe look into / think about `scipy.sparse` like @drammock mentioned?

adam2392: Yeah looks like scipy has a sparse graph representation that supports nans/infs 

https://docs.scipy.org/doc/scipy/reference/sparse.csgraph.html#module-scipy.sparse.csgraph

I'll have to do some reading to see if Xarray plays nicely with that, else this will just be an edge case in the implementation for now.

larsoner:
I don't think we need to insert Nan if it's a sparse representation. Instead, if the element is not in the matrix we assume it's not computed

adam2392:
Unfortunately, scipy sparse would not work I think, since all these connectivity methods place a meaning on the value "0". 

So I'm thinking of making this work at the "container" level. I'm thinking that if `indices` is defined, then that means there is one less "axis" defined in the data array. e.g.

dense all-to-all: `(N, N, freqs, times)`, with names `(node_in, node_out, freqs, times)` would become `(len(indices), freqs, times)`, and then the names would be `(node_in to node_out, freqs, times)` and other metadata that depends on axes would need to also be re-defined appropriately.

larsoner:
> Unfortunately, scipy sparse would not work I think, since all these connectivity methods place a meaning on the value "0".

I don't think this is necessarily a problem because you can have sparse matrices with explicit zeros. Also for ndarray-sparse there is https://github.com/pydata/sparse. But I agree it's not ideal.

> dense all-to-all: (N, N, freqs, times), with names (node_in, node_out, freqs, times) would become (len(indices), freqs, times), and then the names would be (node_in to node_out, freqs, times) and other metadata that depends on axes would need to also be re-defined appropriately.

One unified dense way would be to always have it be:

1. Data of shape `(n_estimated, freqs, times)`
2. attribute `indices`, which is an ndarray of shape `(2, n_estimated)`
3. attribute `node_names`, which is of length `n_nodes`

All index pairs in `indices` must be between `0` and `n_nodes - 1` (inclusive). The all-to-all case can have `indices='all'`, which would be functionally equivalent to `indices=(0, 0), (0, 1), ..., (0, n_nodes-1), (1, 0), (1, 1), ..., (n_nodes-1, n_nodes-2), (n_nodes-1, n_nodes-1))` (but why bother storing it). We could have `indices='symmetric'` which would just store the upper-right triangle (and not bother with the indices, for compactness).

adam2392:
> One unified dense way would be to always have it be:
> 
> 1. Data of shape `(n_estimated, freqs, times)`
> 2. attribute `indices`, which is an ndarray of shape `(2, n_estimated)`
> 3. attribute `node_names`, which is of length `n_nodes`
> 
> All index pairs in `indices` must be between `0` and `n_nodes - 1` (inclusive). The all-to-all case can have `indices='all'`, which would be functionally equivalent to `indices=(0, 0), (0, 1), ..., (0, n_nodes-1), (1, 0), (1, 1), ..., (n_nodes-1, n_nodes-2), (n_nodes-1, n_nodes-1))` (but why bother storing it). We could have `indices='symmetric'` which would just store the upper-right triangle (and not bother with the indices, for compactness).

I like this API sketch, because it can also account for the 'symmetric' case. I suppose then, `indices` is either `all`, `symmetric` or a list of tuples basically with the default being `all`. 

However, in terms of array operations, like shape, dims, etc. should I assume them to be like a "node X node" array? I'm a little hesitant on exposing it as a `(n_nodes*n_nodes, freqs, times)` shaped array, but I could imagine we provide optionality for representing it also as a connectivity matrix in the `get_data()` and other functions.

larsoner:
> However, in terms of array operations, like shape, dims, etc. should I assume them to be like a "node X node" array? I'm a little hesitant on exposing it as a (n_nodes*n_nodes, freqs, times) shaped array, but I could imagine we provide optionality for representing it also as a connectivity matrix in the get_data() and other functions.

Yes internally I would store everything as `(n_estimated, ...)`, even full/all matrices (so `(n_nodes * n_nodes, ...)`. We can provide options in `get_data(..., output='square')` (default `'condensed'` or `'raveled'` or something?) to output the dense/square shaped `(n_nodes, n_nodes, ...)` versions regardless of what `indices` is. (It's just for some examples, like the PSI visual one you're working on, doing `output='full'` might lead to a MemoryError, so it's not recommended in those cases.) We can cross this bridge later, though.

britta-wstnr:
I like the idea of having an option to get either the square or condensed output, that would make individual plotting solutions way easier for the user (given that there are probably more inputs to compute connectivity for than we can account for).

adam2392:
My only qualm that is coming up rn with my changes is how to make things play nicely with xarray. E.g. I'm wondering if xarray is still worth it? And also how should the collapsed "list of node_in -> node_out" be named?

My intuition is saying keep xarray, but come up some better naming, and then just try to wrangle the API to make it work, since xarray brings some niceties to the dimensions + allows storage of "metadata".

larsoner:
Short version of what we discussed off-GH:

1. Store indices, node names, and compact version of xarray (with 0->N-1 labeled "estimated" axis)
2. Provide convenience function to return square-ized xarray with labeled axes
3. Start from basic implementation and start adapting all examples
4. See what methods need to be added
5. Ensure all new methods are unit tested and documented, etc.

doc/whats_new.rst
```diff
@@ -22,29 +22,26 @@ Current
...
-API
-~~~
+Changes when mne-connectivity was part of MNE-Python
+----------------------------------------------------
 
--
+In July, 2021, ``mne.connectivity`` submodule was ported over from the MNE-Python 
+repo into this repository, ``mne-connectivity``. Starting v0.24 of MNE-Python, that sub-module 
+will be deprecated and development will move over into this repository. Starting v0.25 of MNE-Python,
+that sub-module will completely be removed.
```
britta-wstnr, line 44: I would consider changing "sub-module" to `mne.connectivity` in the last two sentences above, just to be very explicit.

examples/mne_inverse_connectivity_spectrum.py
```diff
@@ -66,20 +66,21 @@
...
 sfreq = raw.info['sfreq']  # the sampling frequency
 
-con, freqs, times, n_epochs, n_tapers = spectral_connectivity(
+con = spectral_connectivity(
     label_ts, method='wpli2_debiased', mode='multitaper', sfreq=sfreq,
     fmin=fmin, fmax=fmax, mt_adaptive=True, n_jobs=1)
+freqs = con.freqs
 
-n_rows, n_cols = con.shape[:2]
+n_rows, n_cols = con.get_data(output='full').shape[:2]
```
britta-wstnr, line 74: nice!

mne_connectivity/effective.py
```diff
@@ -86,23 +87,16 @@ def phase_slope_index(data, indices=None, sfreq=2 * np.pi,
...
     Returns
     -------
-    psi : array
-        Computed connectivity measure(s). The shape of each array is either
-        (n_signals, n_signals, n_bands) mode: 'multitaper' or 'fourier'
-        (n_signals, n_signals, n_bands, n_times) mode: 'cwt_morlet'
+    conn : instance of Connectivity
+        Computed connectivity measure(s). Either a
+        ``SpectralConnnectivity``, or ``SpectroTemporalConnectivity``
+        container.The shape of each array is either
```
britta-wstnr, line 93:
```suggestion
        container. The shape of each array is either
```

mne_connectivity/spectral.py
```diff
@@ -639,22 +643,15 @@ def spectral_connectivity(data, method='coh', indices=None, sfreq=2 * np.pi,
     Returns
     -------
     con : array | list of array
-        Computed connectivity measure(s). The shape of each array is either
-        (n_signals, n_signals, n_freqs) mode: 'multitaper' or 'fourier'
-        (n_signals, n_signals, n_freqs, n_times) mode: 'cwt_morlet'
+        Computed connectivity measure(s). Either an instance of
+        ``SpectralConnectivity``, or ``SpectroTemporalConnectivity``.
```
britta-wstnr, line 647:
```suggestion
        ``SpectralConnectivity`` or ``SpectroTemporalConnectivity``.
```

examples/mne_inverse_connectivity_spectrum.py
adam2392: The API is very nice now! At least in my user experience :p

adam2392:
Yay finally everything passed: https://99-369628653-gh.circle-artifacts.com/0/dev/index.html

britta-wstnr:
I looked at the rendered examples and compared the output to the MNE homepage (stable) and the plots all match. Well done on getting it all up and running here!

mne_connectivity/base.py
adam2392: Xarray has officially been integrated... And it's awesome! 

`dims`, `attrs`, `coords` are just nice semantic features of each data axis.

adam2392:
Lmk if you have any qualms or issues with the way things look and I can go ahead and merge this in!

doc/api.rst
```diff
@@ -2,12 +2,40 @@
 API
 ###
 
+This is the reference for classes (``CamelCase`` names) and functions
+(``underscore_case`` names) of MNE-Connectivity, grouped thematically by analysis
+stage. MNE-Connectivity also provides connectivity data structure
+containers for different types of connectivity data, which are
```
britta-wstnr, line 8: You sometimes call the data structures classes (below) and sometimes containers, that could potentially be confusing for novice users. Listing them extra also makes it sound to me like they are different from the classes you list first.

Maybe something like `The data structure classes contain different types of connectivity data and are described below.`?

@larsoner is there some standard way of doing this?  I see MNE does not mentions "data structure" at all and just lists the classes.

mne_connectivity/base.py
```diff
@@ -0,0 +1,525 @@
...
+        between nodes as a (N, N) array. It can be symmetric, or
+        asymmetric (if it is symmetric, storage optimization will
+        occur).
+
+        The underlying data structure is an ``xarray.DataArray``,
+        with a similar API. We provide support for storing
+        connectivity data in a subset of nodes. Thus the underlying
+        data structure instead of a ``(n_nodes_in, n_nodes_out)`` 2D array
+        would be a ``(n_nodes_in * n_nodes_out,)`` raveled 1D array. This
+        allows us to optimize storage also for symmetric connectivity.
```
britta-wstnr, line 39: Should this go into the `Notes` section?

mne_connectivity/base.py
```diff
@@ -0,0 +1,525 @@
...
+                 n_nodes, **kwargs):
+        """Base class container for connectivity data.
+
+        Connectivity data is anything that represents "connections"
+        between nodes as a (N, N) array. It can be symmetric, or
+        asymmetric (if it is symmetric, storage optimization will
+        occur).
+
+        The underlying data structure is an ``xarray.DataArray``,
+        with a similar API. We provide support for storing
```
britta-wstnr, line 35: similar API to what? This is ambiguous to me - I assume `xarray`?

mne_connectivity/base.py
```diff
@@ -0,0 +1,525 @@
...
+            Whether to squeeze the array or not, by default True.
+
+        Returns
+        -------
+        data : np.ndarray
+            The output connectivity data.
+        """
+        if output not in ['raveled', 'full']:
+            raise ValueError(f'Output of data can only be one of '
+                             f'"raveled, "full", not {output}.')
```
britta-wstnr, line 234: This error and where you test for `"symmetric"` vs `"all"` above (line 68) should probably be tested.

britta-wstnr (commented):
Some comments on documentation and tests. Otherwise looks good!

mne_connectivity/base.py
adam2392: Good point! Done.

mne_connectivity/base.py
adam2392: Ooo yeah that's true. I think the symmetric case needs a bit of work. I'm going to take that out for now because we're using xarray and not sure how to best allow for "memory optimization".

doc/api.rst
```diff
@@ -2,12 +2,40 @@
...
+described below.
+
 :py:mod:`mne_connectivity`:
 
 .. automodule:: mne_connectivity
    :no-members:
    :no-inherited-members:
```
britta-wstnr, line 14: Looking at the rendered Documentation, these lines left me a bit puzzled. It seems to show up as:
```
mne_connectivity:

Connectivity MEG, EEG, iEEG data processing.
```

and `mne_connectivity_  just links to the next line. (Looking at the last commit of July 2nd)

larsoner:
CIs are failing, let me know when they're green and I'll take a look!

doc/api.rst
adam2392: True, I think I copied these lines from MNE-realtime/Python. I took these out though since there was no real purpose.

adam2392:
> CIs are failing, let me know when they're green and I'll take a look!

Oops there were some import errors. Kay its good now!

adam2392:
https://109-369628653-gh.circle-artifacts.com/0/dev/index.html

doc/conf.py
```diff
@@ -206,5 +206,5 @@
 # Enable nitpicky mode - which ensures that all references in the docs
 # resolve.
 
-nitpicky = True
-nitpick_ignore = []
+nitpicky = False
```
larsoner, line 209: This is dangerous as it allows bad links to creep in. Why is it necessary?

doc/conf.py
```diff
@@ -206,5 +206,5 @@
...
+nitpick_ignore = [('py:class', 'type')]
```
larsoner, line 210: It would be better to make the second part more specific, it's not clear where this comes from. See for example what we have in MNE where it's more specific (and thus clearer why it's there, and less likely to catch something that shouldn't be caught):

https://github.com/mne-tools/mne-python/blob/0ba06ba6b4dc446146d9b1a3764a697cd4ab16d2/doc/conf.py#L472-L483

doc/whats_new.rst
```diff
@@ -22,29 +22,26 @@ Current
 Changelog
 ~~~~~~~~~
 
+- Adding ``TemporalConnectivity``, ``SpectralConnectivity`` and ``SpectroTemporalConnectivity`` as a data structure to hold connectivity data, by `Adam Li`_ (:gh:`6`)
```
larsoner, line 25: Single backticks here will link to the objects if you have `default_role = 'py:obj'` in `conf.py`

doc/api.rst
```diff
@@ -2,11 +2,32 @@
 API
 ###
 
-:py:mod:`mne_connectivity`:
```
larsoner, line 5: This will make it so that this no longer works:

```
:mod:`mne_connectivity`
```
So it should be restored. You probably also need to restore the `automodule`. If you don't like the text that gets displayed (from `__init__.py`, I think), you can remove it (I think?) or at least make it better.

mne_connectivity/__init__.py
```diff
@@ -6,9 +6,15 @@
 #
 # License: BSD (3-clause)
 
-__version__ = '0.1'
+__version__ = '0.2dev'
```
larsoner, line 9: Not the right way to format this

https://www.python.org/dev/peps/pep-0440/

```suggestion
__version__ = '0.2.dev0'
```

mne_connectivity/envelope.py
```diff
@@ -95,9 +102,12 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
                              % (epoch_data.shape,))
         n_nodes, n_times = epoch_data.shape
         if ei > 0 and n_nodes != corrs[0].shape[0]:
+            print(ei)
+            print(n_nodes, corrs[0].shape)
```
larsoner, line 106: Cruft

mne_connectivity/envelope.py
```diff
@@ -145,8 +155,33 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
             if absolute:
                 corr = np.abs(corr)
             corr = (corr.T + corr) / 2.
+
         corrs.append(corr)
         del corr
 
+    # ravel from 2D connectivity into 1D array
+    corrs = [corr.flatten() for corr in corrs]
```
larsoner, line 163: It seems like this should happen after `fun(corrs)` for backward compat with `combine` as `callable`

mne_connectivity/envelope.py
```diff
@@ -145,8 +155,33 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
...
+    n_epochs = len(corrs)
     corr = fun(corrs)
-    return corr
+
+    # create the connectivity container
+    times = None
+
+    # create epochs axis
+    if combine is not None:
+        corr = corr[np.newaxis, ...]
```
larsoner, line 173: From an API standpoint, I wonder if we should get rid of `combine` as a function argument here and add it as a method of `EpochsTemporalConnectivity` that you can call to combine to obtain a TemporalConnectivity.

And it actually seems like there are some classes missing, like EpochConnectivity and then Connectivity that have no time space or frequency dims.

In other words, I think this should return an instance of EpochConnectivity, and that class should have a method `.combine` that returns a `Connectivity`

mne_connectivity/envelope.py
```diff
@@ -145,8 +155,33 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
...
+
+    # create time axis
+    corr = corr[..., np.newaxis]
+
+    conn = EpochTemporalConnectivity(
+        data=corr,
+        names=names,
+        times=times,
+        method='envelope-correlation',
```
larsoner, line 182:
```suggestion
        method='envelope correlation',
```

mne_connectivity/spectral.py
```diff
@@ -916,19 +920,58 @@ def spectral_connectivity(data, method='coh', indices=None, sfreq=2 * np.pi,
...
+                freqs=freqs,
+                times=times,
+                n_nodes=n_nodes,
+                method=method,
+                spec_method=mode,
+                indices=indices,
+                n_epochs_used=n_epochs,
+                freqs_used=freqs_used,
+                n_tapers=n_tapers
+            )
```
larsoner, line 965: You can DRY this code (and have better/clearer control flow) with:
```
kwargs = dict(data=_con, ...)
if mode in ('multitaper', 'fourier'):
    klass = SpectralConnectivity
    kwargs.update(times_used=times, ...)
    ...
else:
    assert mode == 'cwt_morlet'
    klass = SpectroTemporalConnectivity
    kwargs.update(times=times, ...)
conn_list.append(klass(**kwargs))
```
it makes it immediately clear what is shared (at the beginning) and what is different (in the branches). Also use of if/else rather than if/elif makes it clear that one branch will *always* be entered, and the `assert` just asserts something that we should know from previous checks above -- I think of these like "active comments" in the sense that it both says "this should be the case" and makes sure it is.

mne_connectivity/tests/test_envelope.py
```diff
@@ -38,44 +39,53 @@ def _compute_corrs_orig(data):
...
+
     # using complex data
-    corr = envelope_correlation(data_hilbert)
-    assert_allclose(corr, corr_orig)
+    corr = envelope_correlation(data_hilbert, names=names)
+    assert_allclose(corr.get_data(), corr_orig.flatten())
+
     # using callable
-    corr = envelope_correlation(data_hilbert,
+    corr = envelope_correlation(data_hilbert, names=names,
```
larsoner, line 55: Names should not be obligatory, if not provided it should just use `np.arange(len(signals))`

mne_connectivity/tests/test_spectral.py
```diff
@@ -114,26 +125,40 @@ def test_spectral_connectivity(method, mode):
...
-            assert np.all(con[1, 0, gidx[0]:gidx[1]] < lower_t)
-            assert np.all(con[1, 0, :bidx[0]] < lower_t)
-            assert np.all(con[1, 0, bidx[1]:] < lower_t), \
-                con[1, 0, bidx[1]:].max()
+            assert np.all(con.get_data(output='full')[
+                          1, 0, gidx[0]:gidx[1]] < lower_t)
+            assert np.all(con.get_data(output='full')
+                          [1, 0, :bidx[0]] < lower_t)
+            assert np.all(
+                con.get_data(output='full')[1, 0, bidx[1]:] < lower_t), \
```
larsoner, line 160: use assert_array_less (changing the code anyway, might as well update to newer best practices)

mne_connectivity/tests/test_spectral.py
```diff
@@ -233,7 +277,7 @@ def test_epochs_tmin_tmax(kind):
         assert kind == 'ndarray'
         tmin = 0
         X = data
-    want_times = np.arange(n_times) / sfreq + tmin
+    # want_times = np.arange(n_times) / sfreq + tmin
```
larsoner, line 280: Would be good to restore some form of this if possible. If you want to know why it's necessary you can use blame, which git shows nicely:

https://github.com/mne-tools/mne-python/blame/main/mne/connectivity/tests/test_spectral.py#L236

You can see this was the direct result of a PR:

https://github.com/mne-tools/mne-python/pull/8861

Made in response to a bug. So yes it seems necessary to ensure... all the stuff that PR talks about.

setup.cfg
```diff
@@ -23,7 +23,7 @@ addopts =
     --durations=20 --doctest-modules -ra --cov-report= --tb=short
     --doctest-ignore-import-errors --junit-xml=junit-results.xml
     --ignore=doc --ignore=examples --cov=mne_connectivity
-    --capture=sys
+    # --capture=sys
```
larsoner, line 26: Just delete if you don't want it captured

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+
+        Parameters
+        ----------
+        output : str, optional
+            How to format the output, by default 'raveled', which
+            will represent each connectivity matrix as a
+            ``(n_nodes_in * n_nodes_out, 1)`` list. If 'full', then
+            will return each connectivity matrix as a 2D array.
+        squeeze : bool, optional
+            Whether to squeeze the array or not, by default True.
```
larsoner, line 245: I would get rid of this argument. `np.squeeze` and similar usually cause more harm than good. I think it's better to teach people to understand and be explicit about dimensionality

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+
+        Returns
+        -------
+        data : np.ndarray
+            The output connectivity data.
+        """
+        if output not in ['raveled', 'full']:
+            raise ValueError(f'Output of data can only be one of '
+                             f'"raveled, "full", not {output}.')
```
larsoner, line 254: use the `from mne.utils import _check_option` and
```suggestion
        _check_option('output', output, ['raveled', 'full'])
```

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+    def get_data(self, output='raveled', squeeze=True):
+        """Get connectivity data as a numpy array.
+
+        Parameters
+        ----------
+        output : str, optional
+            How to format the output, by default 'raveled', which
+            will represent each connectivity matrix as a
+            ``(n_nodes_in * n_nodes_out, 1)`` list. If 'full', then
+            will return each connectivity matrix as a 2D array.
```
larsoner, line 243: dense is better than full because eventually we can add "sparse" using pydata/sparse or scipy.sparse. Also the list is 1D so it shouldn't have `, 1` at the end (this is 2D).
```suggestion
            ``(n_nodes_in * n_nodes_out,)`` list. If 'dense', then
            will return each connectivity matrix as a 2D array.
```

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+                if self.is_epoched:
+                    data[:, row_idx, col_idx, ...] = self._data
+                else:
+                    data[row_idx, col_idx, ...] = self._data
+            else:
+                data = self._data.reshape(new_shape)
+
+        if squeeze:
+            return data.squeeze()
+
```
larsoner, line 287:
```suggestion
```

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+        # check that all the channel names are unique
+        if len(names) != len(np.unique(names)):
+            raise ValueError(
+                'New channel names are not unique, renaming failed')
+
+        # rename the new names
+        self._obj.attrs['node_names'] = names
+
+    def plot_circle(self, **kwargs):
```
larsoner, line 333: There are no docs for this function. You probably want `copy_function_doc_to_method_doc` or whatever it's called from MNE

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+        plot_connectivity_circle(
+            self.get_data(),
+            node_names=self.names,
+            indices=self.indices, **kwargs)
+
+    def plot_matrix(self):
```
larsoner, line 339: Don't put these in until they're implemented. If you want to leave them in as an "API plan", comment them out for now. Otherwise when the dev docs deploy these methods will show up as existing, when they don't really (they are no-ops).

mne_connectivity/base.py
```diff
@@ -0,0 +1,548 @@
...
+                         n_epochs_used=n_epochs_used, **kwargs)
+
+    def __repr__(self):  # noqa: D105
+        s = "time : [%f, %f]" % (self.times[0], self.times[-1])
+        s += ", freq : [%f, %f]" % (self.freqs[0], self.freqs[-1])
+        s += ", nave : %d" % self.n_epochs_used
+        s += ', nodes, n_estimated : %d, %d' % (self.n_nodes,
+                                                self.n_estimated_nodes)
+        s += ', ~%s' % (sizeof_fmt(self._size),)
+        return "<SpectroTemporalConnectivity | %s>" % s
```
larsoner, line 445: This seems like it could be DRYed out and put in the base class. If you make each subclass have (private?) class attributes like:
```
class _whateverbase:
    def __repr__(self):
        r = f'<{self.__class__.__name__} '
        if 'times' in self._dims:
            r += ...
        if 'freqs' in self._dims:
            r += ...
        ...
        r += '>'
        return r

class whatever(_whateverbase):
    _dims = ('times', 'freqs')

class whateverelse(_whateverbase):
    _dims = ('times')
```

larsoner (commented):
Looks like it's coming along well! I have some high-level comments that I think would clean up the structure a bit, take a look and see if you think they make sense

doc/whats_new.rst
adam2392: Changed to single ticks.

doc/api.rst
adam2392: Added it back in.

mne_connectivity/envelope.py
adam2392: Wouldn't the "callable" still work on "axis=0"? Since we're just flattening the (N, N) connectivity arrays?

mne_connectivity/envelope.py
adam2392: Getting rid of `combine` makes sense to me, and I think I can do that on the next PR to clean this one up if you don't mind?

Then I have preliminarily added the EpochConnectivity and Connectivity data structures in now.

mne_connectivity/envelope.py
larsoner: We didn't put constraints on how the callable must work, so if for some reason it relied on operating on (N, N) shape then the order you have it in now will break it

mne_connectivity/envelope.py
larsoner: Yes I can live with it in a subsequent PR

mne_connectivity/envelope.py
adam2392: Okay that makes sense! Fixed the order

doc/conf.py
adam2392: Added these back in. I think I wasn't sure how to get things like "`SpectralConnectivity`" to actually find the object. I was doing something like

:class:`SpectralConnectivity`.

larsoner: even
```
:class:`SpectralConnectivity`
```
will work so long as whatever is being documented at the time is documented in the API docs at the same level as `SpectralConnectivity` (e.g., if you're putting this ref in `mne_connectivity.phase_slope_index` function and the class is documented as `mne_connectivity.SpectralConnectivity` and hence at the same `mne_connectivity` level). If you remove this and have problems let me know and I can probably help find a solution that links properly

larsoner:
@adam2392 something must be wrong with your local `git` configuration, the commits you're pushing are not properly linking to your GitHub account (see the icons to the left of your commits toward the end of this page, they should link to your GitHub avatar but don't). Looking at the `git log` locally I see:
```
Author: Adam Li <adam2392@Adams-MacBook-Pro-2.local>
```
so you might need to edit `~/.gitconfig` to have:
```
[user]
        name = ...
        email = ...
```
set properly. `git config --global` can probably be used to do the same...

adam2392:
https://118-369628653-gh.circle-artifacts.com/0/dev/index.html

adam2392:
> @adam2392 something must be wrong with your local `git` configuration, the commits you're pushing are not properly linking to your GitHub account (see the icons to the left of your commits toward the end of this page, they should link to your GitHub avatar but don't). Looking at the `git log` locally I see:
> 
> ```
> Author: Adam Li <adam2392@Adams-MacBook-Pro-2.local>
> ```
> 
> so you might need to edit `~/.gitconfig` to have:
> 
> ```
> [user]
>         name = ...
>         email = ...
> ```
> 
> set properly. `git config --global` can probably be used to do the same...

Yay it's fixed!

adam2392:
Cool, I think this is good to go now! 

I'll address

`symmetric`, example of Connectivity, refactoring `envelope_correlation` in the next PR.

PR #27, opened 2021-08-04
[MRG] Adding tests for viz.

adam2392:
PR Description
--------------

Adds tests for viz that weren't added before.

Moved back `circular_layout` to mne.viz according to @larsoner comments.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


britta-wstnr:
If we want to follow MNE-Python logic, these should live in ` mne_connectivity/viz/tests` and not `mne_connectivity/tests/viz`.

adam2392:
> If we want to follow MNE-Python logic, these should live in ` mne_connectivity/viz/tests` and not `mne_connectivity/tests/viz`.

Done.

mne_connectivity/viz/tests/test_circle.py
```diff
@@ -0,0 +1,91 @@
...
+    con = np.random.RandomState(0).randn(68, 68)
+    plot_connectivity_circle(con, label_names, n_lines=300,
+                             node_angles=node_angles, title='test',
+                             )
+
+    pytest.raises(ValueError, circular_layout, label_names, node_order,
+                  group_boundaries=[-1])
+    pytest.raises(ValueError, circular_layout, label_names, node_order,
+                  group_boundaries=[20, 0])
+    plt.close('all')
```
larsoner, line 91: `plt.close('all'` should not be necessary if you have the same `autouse` fixture that MNE-Python has that calls `plt.close('all')` automagically

adam2392: I think I copy pasted this directly from MNE-Python, but let me take a look rn.

larsoner: Yeah we still have a lot left in the MNE-Python codebase that can be removed

larsoner: ... you appear not to have a `conftest.py` (?), so you might want to make one and copy the `autouse` fixture from MNE-Python from it

britta-wstnr (commented):
Looks good to me, but can you check your failing tests?

adam2392:
Not sure why the GH actions were not running to completion, so just restarted the jobs.

mne_connectivity/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+#
+# License: BSD-3-Clause
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def close_all():
+    """Close all matplotlib plots, regardless of test status."""
+    # This adds < 1 µS in local testing, and we have ~2500 tests, so ~2 ms max
```
larsoner, line 12: This comment should be removed as it's not correct for this repo

larsoner (approved):
LGTM +1 for merge even without the comment removal

mne_connectivity/conftest.py
adam2392, line 12:
```suggestion
```

adam2392:
@larsoner actually it looks almost like the entire unit testing suite is stalling on the GH Ubuntu build. Is there anything special with the conftest that I'm missing? 

My suspicion was perhaps the plots are not closing inside the GH actions, which makes them time out?

larsoner:
Oh yes you'll need the other viz fixture in MNE-Python that sets the backend to agg so that nothing blocks (the Qt backend, which is default on most OSes, will block):

https://github.com/mne-tools/mne-python/blob/main/mne/conftest.py#L170-L210

You shouldn't need the `traits` stuff but the exception handling is useful

adam2392:
Hmmm the viz is still stalling now without the explicit `plt.close()`?

mne_connectivity/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    # users might have changed locally (but should not otherwise affect
+    # functionality)
+    plt.ioff()
+    plt.rcParams['figure.dpi'] = 100
+    try:
+        from traits.etsconfig.api import ETSConfig
+    except Exception:
+        pass
+    else:
+        ETSConfig.toolkit = 'qt4'
```
larsoner, line 50: Delete, you shouldn't need this

larsoner (commented):
Based on where it's dying I don't think `matplotlib` is the problem but rather VTK/Mayavi, probably due to `vtk:           9.0.3`. VTK <= 9.0.1 should be used (see how we do it in MNE if you want to follow our patterns). Also the `renderer` tests are not running with `pyvistaqt` it looks like...

larsoner:
@adam2392 FYI it's better practice to open PRs from your fork rather than the main repo

adam2392:
Tbh, I'm not sure why things are still stalling:

- I matched the `requirements.txt` file at :https://github.com/mne-tools/mne-python/blob/main/requirements.txt
- I think I have most of the `tools/*.sh` files match what we have in this repo
- I have the `- run: ./tools/setup_xvfb.sh` in my GH action workflow.
- copied over information in the `mne/conftest.py` that was relevant to plotting 

Did I miss something?

larsoner:
I think VTK 9.0.3 is the problem, we don't actually test it over in MNE-Python. For all builds we use conda except:

1. The Linux 3.9 pre one , and for that we install a 9.0.1dev wheel that I compiled for Linux
2. The Windows 3.8 pre where it installs 9.0.1

I pushed a commit that might fix it...

larsoner:
It's a bit annoying that the GH actions are taking forever, so I added auto-cancelling via concurrency (see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency)

larsoner:
1. Added `pytest-timeout` to hopefully help with timeouts
2. Added `pytest_configure` to run `matplotlib_config` automagically
3. Added warnings-as-errors
4. Cast to real in VAR code to avoid a warning about that being done (comes from the `sqrtm` call)

@adam2392 can you see if this all makes sense to you?

larsoner:
Bumped to Python 3.7+ requirement (same as MNE) and got rid of 3.6 on CIs

adam2392:
Ah nice, was the issue then coming from the fact that it was Python3.6?, or that Python3.9 was too high for certain vtk versions?

Or was it that I didn't set the matplotlib_config correctly before?

The LOC makes sense, but jw for my own sake.

larsoner:
> Or was it that I didn't set the matplotlib_config correctly before?

`matplotlib_config` was not being called because it wasn't being set to being automatically used (now it is by `pytest_configure`), and hence it was getting stuck when a matplotlib figure was created and shown.

The VTK stuff probably does not matter, I'll try setting that back.

adam2392:
Thanks @larsoner !

PR #31, opened 2021-08-18
[MRG] Adding notes to readme and spectral_connectivity

adam2392:
PR Description
--------------

Describes differences in connectivity measures.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,15 +89,17 @@ https://mne.discourse.group
 A Note About Connectivity
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In the neuroscience community as of 2021, the term "connectivity" can have many
-different meanings. There is the common question of whether or not
-"connectivity" detected via some measure actually implies a connection between
-two brain regions. That is, is the connectivity causal? Even if one calls their
-connectivity measure, "causal", it depends on the data and the underlying
-assumptions. Some common interpretations of connectivity are in are in
-`Schoffelen`_.
+In the neuroscience community as of 2021, the term "functional connectivity" can 
+have many different meanings and comprises many different measures. Some of 
+these measures are directed (i.e. try to map a statistical causal relationship between
+brain regions), others are non-directed. Please note that the interpretation of your 
+functional connectivity measure depends on the data and underlying
+assumptions. 
+For a taxonomy of functional connectivity measures and information on the 
+interpretation of those measures, we refer to
+`Bastos and Schoffelen`_.
 
 In mne-connectivity, we do not claim that any of our measures imply causal
 connectivity.
 
-.. _Schoffelen: https://pubmed.ncbi.nlm.nih.gov/26778976/
+.. _Bastos and Schoffelen: https://pubmed.ncbi.nlm.nih.gov/26778976/
--- a/README.rst
+++ b/README.rst
@@ -90,17 +90,14 @@ A Note About Connectivity
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 In the neuroscience community as of 2021, the term "connectivity" can have many
-different meanings. Some common interpretations are:
+different meanings. There is the common question of whether or not
+"connectivity" detected via some measure actually implies a connection between
+two brain regions. That is, is the connectivity causal? Even if one calls their
+connectivity measure, "causal", it depends on the data and the underlying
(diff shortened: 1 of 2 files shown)


larsoner (approved):
LGTM, but let's wait to see if @britta-wstnr agrees

britta-wstnr (commented):
See comments for my view on the "connectivity issue".

Independently of this: I wonder if we should list the possible methods in the docstring under `method`? The notes section is really long now and a list might benefit those users who know which measure they want to compute but cannot remember whether it was `pli2unbiased` or `pli2_unbiased`, for example.

mne_connectivity/spectral.py
```diff
@@ -662,15 +662,15 @@ def spectral_connectivity(data, names=None, method='coh', indices=None,
...
-    possible.
-
-    These measures are not to be intepreted over time. Each Epoch passed into
-    the dataset is intepreted as an independent sample of the same connectivity
-    structure. Within each Epoch, it is assumed that the spectral measure is
-    stationary. **Thus, spectral measures computed with only one Epoch will result
-    in errorful values.**
+    these to "brain connectivity" results by thresholding values below a
+    certain number. This should be done through some cross-validation procedure
+    if possible.
```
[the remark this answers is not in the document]
adam2392: Okay that's a good point, I can add some discussion there on what you wrote.

Regarding what I wrote originally: I would say that is a special case then of directed/non-directed measure of connectivity, where the thresholding step is analogous to an alpha value in hypothesis testing, which controls for false negatives. Thus would we agree then it is a notion of computing "connectivity", but just one specific model?

It sounds like (and I agree) that you would like me though to generalize the text here, which I will do.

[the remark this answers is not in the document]
britta-wstnr: Yes, I would leave this out here and rather discuss in a tutorial about statistics. The way it is written now, I also am not sure how much information it would give a new user: questions that pop up are e.g. thresholding how, is not every "brain measure" a pure estimate (even when statistically controlled), etc.
I think we rather want to acknowledge that the opinions on "what can be called connectivity" might differ - and converge on one terminology that we keep throughout the toolbox as to not confuse users further.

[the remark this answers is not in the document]
adam2392: Okay sounds good. took it out.

README.rst
```diff
@@ -85,3 +85,22 @@ Forum
...
+- "actual" connectivity as a graph, where edge values indicate a "connection", and an edge value of zero indicates "no connection"
+
+In the graph interpretation, this can be further stratified into:
+
+- directed, or undirected graph (symmetric, or asymmetric)
+- weighted, or unweighted graph (values of 0 and 1 vs continuous values)
+
+The functions implemented in mne-connectivity generally imply the "all-to-all"
+values case. This is an important distinction because the values between each node
+computed from any of these functions do not necessarily imply a connection. These are
```
britta-wstnr, line 105:
```suggestion
computed from any of these functions do not necessarily imply a causal connection. These are
```

README.rst
```diff
@@ -85,3 +85,22 @@ Forum
...
+different meanings. Some common interpretations are:
+
+- all-to-all values computed from some bivariate or multivariate function over signals
+- "actual" connectivity as a graph, where edge values indicate a "connection", and an edge value of zero indicates "no connection"
+
+In the graph interpretation, this can be further stratified into:
+
+- directed, or undirected graph (symmetric, or asymmetric)
+- weighted, or unweighted graph (values of 0 and 1 vs continuous values)
+
```
britta-wstnr, line 102: Mh, I am not sure if your point 1 (all-to-all bivariate/multivariate function) is not the same as point 3 (directed). Or in other words: the distinction of directed and undirected as well as weighted and unweighted is not exclusive to graph measures.

I wonder if we should just subscribe to a published opinion on this, WDYT @larsoner ? 
There is the tutorial paper I mentioned before by _Bastos & Schoffelen 2016_ which have a IMO good taxonomy on connectivity measures.

larsoner: I am not confident in the correct taxonomy, so I agree that having / following some paper's definitions is the way to go

adam2392: Okay added a link to the paper and removed this text. Can save it for a statistics tutorial!

PR #33, opened 2021-08-18
[MRG] Adding connectivity class example.

adam2392:
PR Description
--------------

Adds a connectivity class example demonstrating basic usage of the different properties of the classes.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+
+==============================
+Using the connectivity classes
+==============================
+
+Compute different connectivity measures and then demonstrate
+the utility of the class.
+
+Here we compute the Phase Lag Index (PLI) between all gradiometers and showcase
+how we might interface with the connectivity class.
```
britta-wstnr, line 12:
```suggestion
how we can interact with the connectivity class.
```
Maybe?

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+# Pick MEG gradiometers
+picks = mne.pick_types(raw.info, meg='grad', eeg=False, stim=False, eog=True,
+                       exclude='bads')
+
+# Create epochs for the visual condition
+event_id, tmin, tmax = 3, -0.2, 1.5  # need a long enough epoch for 5 cycles
+epochs = mne.Epochs(raw, events, event_id, tmin, tmax, picks=picks,
+                    baseline=(None, 0), reject=dict(grad=4000e-13, eog=150e-6))
+
+# Compute connectivity for band containing the evoked response.
```
britta-wstnr, line 45:
```suggestion
# Compute connectivity for the band that contains the evoked response (4-9 Hz).
```

Generally, it would of course be nicer to use some genuine oscillatory activity. There might be alpha desynchronization going on here and I think we show gamma activity in one example of MNE-Python as well?

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+
+# the coordinates for each axis of the data
+print(con.coords)
+
+# the underlying data is stored "raveled"
+print(con.shape)
+
+# however, if one asks for the output dense data, then
+# the shape will show the N by N connectivity
+print(con.get_data(output='dense').shape)
```
britta-wstnr, line 70: Maybe this would be a good place to also explain the difference between those representations and mention when one would prefer to use which output?

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+
+# the number of nodes matches the number of signals
+# used to compute the spectral measure
```
britta-wstnr, line 73: I find this framing confusing. "Signals" could also refer to the observations, i.e. epochs. I would prefer to make it explicit, so for example say "electrodes" instead of "signals" or "signals from electrodes" if you must keep signal.

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+print(con.n_nodes)
+
+# the names of each node correspond to the electrodes
```
britta-wstnr, line 76:
```suggestion
# the names of each node correspond to the electrode names
```

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+print(con.names)
+
+# The underlying data is stored as an xarray, so we have access
+# to DataArray attributes. You can store metadata relevant to the
+# estimated measure, or other relevant metadata.
+print(con.attrs)
```
britta-wstnr, line 82: Either here or in an upcoming example it would be great to show how this can be done with a little example.

examples/connectivity_classes.py
```diff
@@ -0,0 +1,96 @@
...
+                    baseline=(None, 0), reject=dict(grad=4000e-13, eog=150e-6))
+
+# Compute connectivity for band containing the evoked response.
+# We exclude the baseline period:
+fmin, fmax = 4., 9.
+sfreq = raw.info['sfreq']  # the sampling frequency
+tmin = 0.0  # exclude the baseline period
+epochs.load_data().pick_types(meg='grad')  # just keep MEG and no EOG now
+con = spectral_connectivity(
+    epochs, method='pli', mode='cwt_morlet', sfreq=sfreq, fmin=fmin, fmax=fmax,
```
britta-wstnr, line 52: This call fails, you'd have to use `cwt_freqs`.

examples/connectivity_classes.py
adam2392: Added lmk what you think?

examples/connectivity_classes.py
adam2392: I added some additional details. Most of the storage occurs under the hood.

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+
+# the underlying data is stored "raveled"
+print(con.shape)
+
+# However, if one asks for the output dense data, then the shape will show the
+# N by N connectivity. In general, you might prefer the raveled version if you
+# specify a subset of indices (e.g. some subset of sources) to compute
+# bivariate connectivity over, or if you have a symmetric measure
+# (e.g. coherence). The 'dense' output is nice to obtain an actual square
+# matrix, which can be used for post-hoc analysis that expects a matrix shape.
```
britta-wstnr, line 77: I would make this a new paragraph and text instead of a comment. (i.e., add a `# %%` and maybe give it a subtitle. I suspect many users wanting to check especially this part out, with a title it's easy to find.)

britta-wstnr: Maybe start out by explaining that the data can be stored in two ways (you say "data is stored as raveled" above, so now you can start the sub-section by saying, there are two ways, "raveled" and "dense")

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+# the dimensions of the data corresponding to each axis
+print(con.dims)
+
+# the coordinates for each axis of the data
+print(con.coords)
+
+# the underlying data is stored "raveled"
+print(con.shape)
+
+# However, if one asks for the output dense data, then the shape will show the
```
britta-wstnr, line 72:
```suggestion
# However, if one asks for the output dense data, then the data will represent the full
```

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+# N by N connectivity. In general, you might prefer the raveled version if you
+# specify a subset of indices (e.g. some subset of sources) to compute
+# bivariate connectivity over, or if you have a symmetric measure
```
britta-wstnr, line 75:
```suggestion
# specify a subset of indices (e.g. some subset of sources) for the computation
# of a bivariate connectivity measure or if you have a symmetric measure
```

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+# (e.g. coherence). The 'dense' output is nice to obtain an actual square
```
britta-wstnr, line 76:
```suggestion
# (e.g. coherence). The 'dense' output on the other hand provides an actual square
```

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+
+# the names of each node correspond to the electrode names
+print(con.names)
+
+# The underlying data is stored as an xarray, so we have access
+# to DataArray attributes. You can store metadata relevant to the
+# estimated measure, or other relevant metadata. For example,
+# the method name used to compute this connectivity is the 'pli' measure.
+print(con.attrs)
+print(con.attrs.get('method'))
```
britta-wstnr, line 92: I would also make this a new subsection and give it a sub-title.

examples/connectivity_classes.py
```diff
@@ -0,0 +1,106 @@
...
+
+# %%
+# Other properties of the connectivity class, special for
+# the spectro temporal connectivity.
```
britta-wstnr, line 96:
```suggestion
# Other properties of the connectivity class, special to
# the spectro-temporal connectivity class.
```

britta-wstnr: Not sure about the "to" - not a native speaker ;-)

examples/connectivity_classes.py
britta-wstnr: I was wondering if it would be possible to show how you would "store metadata" - as you write it here, it seems I could store, let's say, the information that this estimate was done on pre-processed data, or other metadata I want to contain? 
If we can only store metadata that is automatically stored by our code, your phrasing of "You can store metadata" might be confusing.

adam2392: Ah I see, okay added an example!

PR #36, opened 2021-09-03
ENH: Add symmetric orthogonalization

larsoner:
This implements symmetric orthogonalization following https://www.sciencedirect.com/science/article/pii/S1053811915002670.

@agramfort @dengemann @SherazKhan any thoughts on the utility of this versus the already-implemented pairwise orthogonalization? It seems to have the limitation of requiring n_labels < rank, but some advantages mentioned in the paper above (better sensitivity / fewer false alarms among them).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,16 +99,16 @@ jobs:
         # Load our data
         - restore_cache:
             keys:
-              - data-cache
+              - data-cache-sample
+
+        # Load our data
+        - restore_cache:
+            keys:
+              - data-cache-bst-resting
 
         - run:
             name: Check PyQt5
             command: LD_DEBUG=libs python -c "from PyQt5.QtWidgets import QApplication, QWidget; app = QApplication([])"
-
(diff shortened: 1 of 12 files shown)


dengemann:
@larsoner cool! Once it is implemented we should try to understand its benefits :)

larsoner:
> Once it is implemented we should try to understand its benefits :)

It's implemented here, feel free to try it out on some data

agramfort (commented):
yes this technique as a constraints on the rank of the data. I think it's still a useful and accepted approach in the community.

I have not checked the numerics.

SherazKhan:
This method is indeed valid and has been used in multiple application papers.

larsoner:
https://253-369628653-gh.circle-artifacts.com/0/dev/auto_examples/mne_inverse_envelope_correlation.html

agramfort (approved):

adam2392 (approved):

adam2392 (approved):

PR #52, opened 2021-11-04
FIX: No more _import_fft

larsoner:


larsoner:
This one is straightforward so I'll merge. @adam2392 it would be great to have a release to fix this since it's going to break for anyone using 1.0.dev0 of MNE, but no huge rush because for now for MNE I can have CircleCI use the `main` branch of `mne-connectivity`

PR #89, opened 2022-03-17
Fix code-block

alexrockhill:
Typo I missed

### DIFF
--- a/mne_connectivity/viz/circle.py
+++ b/mne_connectivity/viz/circle.py
@@ -141,32 +141,15 @@ def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
                 subplot = (subplot,)
             ax = plt.subplot(*subplot, polar=True)
 
-    try:
-        return _plot_connectivity_circle(
-            con=con, node_names=node_names, indices=indices, n_lines=n_lines,
-            node_angles=node_angles, node_width=node_width,
-            node_height=node_height, node_colors=node_colors,
-            facecolor=facecolor, textcolor=textcolor,
-            node_edgecolor=node_edgecolor, linewidth=linewidth,
-            colormap=colormap, vmin=vmin, vmax=vmax, colorbar=colorbar,
-            title=title, colorbar_size=colorbar_size,
-            colorbar_pos=colorbar_pos, fontsize_title=fontsize_title,
-            fontsize_names=fontsize_names, fontsize_colorbar=fontsize_colorbar,
-            padding=padding, ax=ax, interactive=interactive,
-            node_linewidth=node_linewidth, show=show)
-    except TypeError as err:
-        if len(err.args) != 1 or err.args[0] not in (
-            "cannot unpack non-iterable NoneType object",
-                "unsupported operand type(s) for -: 'int' and 'NoneType'"):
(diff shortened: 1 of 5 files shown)


alexrockhill:
Ok done messing with stuff, this is really all the fixes

alexrockhill:
So the colorbar is all messed up because https://github.com/mne-tools/mne-python/pull/10443 needs to stop the colorbar from being removed from gridspec. That can be added back after the 1.0 release. Right now on `main`, the title overlaps so it's really not great either way, will be totally fixed soon.

adam2392:
Looks like circleCI is failing.

alexrockhill:
> Looks like circleCI is failing.

Sorry! It looks like it was an extra space `#  .. code-block` not `# .. code-block`, dang.

Okay, once https://github.com/mne-tools/mne-python/pull/10443 merges, there will be errors if `colorbar_scale` and `colorbar_pos` are `None` so I just wrapped that in a try-catch and changed them to `None` defaults. That is nice for maintainability so that we don't have to change those, and I'm not sure the rationale for overriding the matplotlib defaults. We could keep the small colorbar as a default but it's much better not to have `gridspec=False` for the colorbar so that it doesn't clash with the plots when you try and `subplots_adjust` or `tight_layout`.

mne_connectivity/viz/circle.py
```diff
@@ -141,14 +141,32 @@ def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
...
-        node_angles=node_angles, node_width=node_width,
-        node_height=node_height, node_colors=node_colors,
-        facecolor=facecolor, textcolor=textcolor,
-        node_edgecolor=node_edgecolor, linewidth=linewidth,
-        colormap=colormap, vmin=vmin, vmax=vmax, colorbar=colorbar,
-        title=title, colorbar_size=colorbar_size, colorbar_pos=colorbar_pos,
-        fontsize_title=fontsize_title, fontsize_names=fontsize_names,
-        fontsize_colorbar=fontsize_colorbar, padding=padding, ax=ax,
-        interactive=interactive, node_linewidth=node_linewidth, show=show)
+    try:
```
agramfort, line 144: this is really a code smell. Please avoid try/except blocks like this and try to catch errors explicitly as early as possible

alexrockhill: The issue is that the code is in `mne-python` and there is no way to undo the default arguments since they are passed since if you pass them to `matplotlib` they will throw exceptions for `None`.

Maybe I will just leave the small colorbar as default in the other PR but just remove it from gridspec and then we don't have to do this.

alexrockhill:
Ok, I think I was overthinking it, this just fixes those tiny leftovers and hopefully I did the `code-block` right this time

agramfort (approved):

alexrockhill:
The colorbar is still overlapped with the plot here https://508-369628653-gh.circle-artifacts.com/0/dev/auto_examples/mixed_source_space_connectivity.html#sphx-glr-auto-examples-mixed-source-space-connectivity-py but that will be fixed when https://github.com/mne-tools/mne-python/pull/10443 merges.

PR #100, opened 2022-06-14
[GSOC] implement example of state-space model for connectivity

### LINKED ISSUE #99: Gathering feedback on API Proposal for state-space model
As part of the Google Summer of Code program, I am constructing an [API proposal](https://docs.google.com/document/d/1N50QP1cCTixOb8OddOJydffrv2Tv2-Zc2I5Hs_V4Sa8/edit?usp=sharing) to implement a state-space model using a Kalman Filter and an Expectation Maximization algorithm to estimate functional connectivity between regions of interest as recorded in M/EEG. The proposal also outlines the implementation of methods to compute the statistical significance between functionally connected networks.

This is my first API proposal so any feedback from variables to proposal structure is sincerely appreciated!

jadrew43:
# PR Description

## Google Summer of Code (2022) project 

Closes #99 

WIP: Linear Dynamic System (state-space model using EM algorithm to find autoregressive coefficients) to infer functional connectivity by interpreting autoregressive coefficients as connectivity strength. The model uses M/EEG data as input, and outputs time-varying autoregressive coefficients for source space labels.

### Completed during GSoC

- [x] A user-friendly API that allows the user to work easily with MEG and/or EEG EEG data, following MNE-Python's local standards and conventions for usability as much as possible:
    - Most of the code complexity is hidden from the user in the backend for the simplest interface:
    
```
data_path = mne.datasets.sample.data_path()

raw = mne.io.read_raw_fif(raw_fname).crop(tmax=60)
events = mne.find_events(raw, ...)

event_dict = {...}
epochs = mne.Epochs(raw, events, tmin=-0.2, tmax=0.7, event_id=event_dict,
                    preload=True).pick_types(meg=True,eeg=True)


fwd_fname = sample_folder / '....'
fwd = mne.read_forward_solution(fwd_fname)

cov_fname = sample_folder / 'sample_audvis-cov.fif'
cov = mne.read_cov(cov_fname)

label_names = ['Aud-lh', 'Aud-rh', 'Vis-lh', 'Vis-rh']
labels = [mne.read_label(sample_folder / 'labels' / f'{label}.label',
                          subject='sample') for label in label_names]

model = LDS(lam0=0, lam1=100)
model.add_subject('sample', condition, epochs, labels, fwd, cov)
model.fit(...)
model.fit()
At = model.A
assert At.shape == (len(labels), len(labels), len(epochs.times))
```

- [x] Preprocess it to a format meant to increase the SNR of the data
- [x] Downsample the data for faster processing.
- [x] Utilizes forward and covariance matrices in the API well as the labels for the regions of interest (ROIs) of the dataset
- [x] PCA is used to reduce the dimensionality of the data
- [x] `model.fit` using the Expectation Maximization algorithm to fit the autoregressive coefficients of the state-space model, mapping the sensor data to each ROI, and computing the connectivity strength between ROI pairs
- [x] Plotting of the time-varying coefficients in a matrix format to observe the strength of connection between each pair of ROIs
- [x] An example script showing how to use the function and interpret its outputs using the MNE-Python `sample` dataset
- [x] Basic unit tests have been written and partially incorporated

Check-out this [link](https://blogs.python-gsoc.org/en/jadrew43s-blog/) to see my weekly progress. All of the code in this PR is new to MNE-Python's repositories. 

### Todo after GSoC

- [ ] Finish unit tests
- [ ] Finish replacing redundant functionality with MNE-Python equivalents (e.g., `scale_data` with `compute_whitener` and dot products)
- [ ] The current implementation uses `autograd` as a dependency, which is no longer actively developed (but still maintained). The code should be updated to incorporate `JAX`, which includes the features of `autograd` and is being actively developed (`autograd_linalg` should be replaced by `scipy.linalg`).

## Merge checklist

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


adam2392 (commented):
Left some minor suggestions

Lmk when you want us to activate the CI pipelines + do a more in-depth look at the code!

examples/mne_util.py
```diff
@@ -0,0 +1,294 @@
+""" MNE-Python utility functions for preprocessing data and constructing
+    matrices necessary for MEGLDS analysis """
+
+from __future__ import division
+from __future__ import print_function
+from __future__ import absolute_import
```
adam2392, line 6: Not needed since we'll use Python3.7+

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,63 @@
...
+                    preload=True)
+# del raw
+
+## compute covariance 
+noise_cov = mne.compute_covariance(epochs, tmax=0)
+labels = mne.read_labels_from_annot('sample', subjects_dir=subjects_dir)
+roi_to_src = ROIToSourceMap(fwd, labels) # compute ROI-to-source map
+scales = {'eeg_scale' : 1, 'mag_scale' : 1, 'grad_scale' : 1}
+fwd_sr_sn, fwd_roi_sn, snsr_cov, epochs = \
+        scale_sensor_data(epochs, fwd, noise_cov, roi_to_src, **scales)
```
larsoner, line 54: To me the `ROIToSourceMap` and `scale_sensor_data` should just be hidden steps in the `model.fit` step, no? Does a user really need access to these?

This would allow `MEGLDS` (we should probably rename it just `LDS` or something since it's not MEG-specific at all) to have all the properties of ROIToSourceMap internally

examples/mne_util.py
```diff
@@ -0,0 +1,294 @@
...
+        if isinstance(label, mne.Label):
+            return 0 if label.hemi == 'lh' else 1
+        elif isinstance(label, mne.BiHemiLabel):
+            hemis = [None] * 2
+            for i, lab in enumerate([label.lh, label.rh]):
+                hemis[i] = 0 if lab.hemi == 'lh' else 1
+            return hemis
+
+
+def morph_labels(labels, subject_to, subjects_dir=None):
```
larsoner, line 151: Let's remove this for now, does not seem relevant/needed

examples/mne_util.py
```diff
@@ -0,0 +1,294 @@
...
+
+    Q = cov.data
+    if not np.allclose(np.dot(proj, Q), Q):
+        Q = np.dot(proj, np.dot(Q, proj.T))
+        cov.data = Q
+
+    return fwd, cov
+
+
+def scale_sensor_data(epochs, fwd, cov, roi_to_src, eeg_scale=1., mag_scale=1.,
```
larsoner, line 207: We probably don't really need this function. Or at least it doesn't need to be public. In other words, it should have an underscore prefix `_` prepended to the name to make it clear that it's not intended for users to ever see or use it.

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Jun 14 13:15:43 2022
+
+@author: jordandrew
+
+For 'mne-connectivity/examples/' to show usage of MEGLDS 
+Use MNE-sample-data for auditory/left
```
larsoner, line 9: See how we construct examples/tutorials with proper titles etc.

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+
+## import necessary libraries
+import mne
+import numpy as np 
+import matplotlib.pyplot as plt
+from mne_util import ROIToSourceMap, scale_sensor_data #mne_util is from MEGLDS repo
+
+## define paths to sample data
+data_path = '/Users/jordandrew/Documents/MEG/mne_data/MNE-sample-data'
```
larsoner, line 19: See how we do it in other examples/tutorials in MNE-Python using `data_path`

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,63 @@
...
+event_dict = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3,
+              'visual/right': 4, 'face': 5, 'buttonpress': 32}
+epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7, event_id=event_dict,
+                    preload=True)
+# del raw
+
+## compute covariance 
+noise_cov = mne.compute_covariance(epochs, tmax=0)
+labels = mne.read_labels_from_annot('sample', subjects_dir=subjects_dir)
+roi_to_src = ROIToSourceMap(fwd, labels) # compute ROI-to-source map
```
larsoner, line 51: To me this is where it should be `model = LDS(fwd, labels, noise_cov, epochs, subjects_dir)`. Or if we're planning on the model being fit over multiple subjects:
```
model = LDS(..., subjects_dir=None)
model.add_subject(subject, fwd, labels, noise_cov, epochs)
model.fit()
```
and at the `add_subject` stage it could do all the whitening/rescaling etc. And at the `.fit` stage it does all the heavy lifting. And the `...` above could be all the stuff about lambdas, `random_state` support, etc. We'd have to see what the minimum set of parameters is we need for the `LDS` class...

larsoner (commented):
Looks like a good start, let's keep iterating here until we're happy with how the interface looks in the example!

cbrnr:
Hello! Just wanted to say hi to include myself in the loop 😄. Could someone quickly explain what the aim of this PR is? Is it to add VAR model based connectivity estimates? State-space sounds like it's implemented as a Kalman filter. At the risk of sounding repetitive, but could we look at https://github.com/scot-dev/scot to see if anything could be re-used here? I've implemented least squares VAR estimation (optionally with regularization) to compute several popular (directed) connectivity measures.

examples/mne_util.py
```diff
@@ -0,0 +1,294 @@
...
+        self.Q_J = Q_J
+        self.offsets = offsets
+        self.n_lhverts = n_lhverts
+        self.n_rhverts = n_rhverts
+        self.labels = labels
+
+        return
+
+    @property
+    def G(self):
```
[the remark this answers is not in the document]
jadrew43: should this be in the docstring of the example script of of `mne_util.py`?

examples/state_space_connectivity.py
jadrew43: The docstring sections on tutorials seem to just be authors and a statement about the license. Happy to edit the author statement, but unsure if I should include "License BSD (3-Clause))" like in other tutorials. Thoughts?

examples/state_space_connectivity.py
jadrew43: I have the normal definition for `data_path` (as defined in tutorials) commented out on the next line, I just use this^^ definition so it doesn't keep downloading to my machine.

examples/mne_util.py
jadrew43: Is this just in reference to `def morph_labels` or the entirety of lines 91-151? I'm still getting used to reading Github.

jadrew43:
Hi @cbrnr you are correct that this PR aims to implement a Kalman filter using an AR model to measure connectivity. Reviewing SCoT is still on my to-do list, thanks for the reminder.

examples/mne_util.py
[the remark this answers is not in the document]
drammock: no. The docstring of `class ROIToSourceMap` should document the parameters of its `__init__` method, and you could include a `Notes` section there that connected the (explicit/descriptive) method and property names to the (brief/opaque) variable names used in the literature. Something like:

```
Notes
-----

- The property ``whatever`` corresponds to the matrix ``G`` in :footcite:`BibtexKey`
```

examples/state_space_connectivity.py
drammock: license is implied / specified at the repo level, no need to include it in a tutorial doc.

examples/state_space_connectivity.py
drammock: that's not how `data_path` works :)  it's smart enough to look to see if you already have the dataset downloaded; it won't re-download every time.

jadrew43: It looks like it wants to download to a different folder than where I keep this project. Can I change the default download path?

adam2392: If you need to work with a local copy because you're modifying the data files multiple times, then you can specify the `path` argument.

See: https://mne.tools/stable/generated/mne.datasets.sample.data_path.html

If you're not touching the data files, I would just leave as is actually cuz you can easily access the path.

jadrew43: Yup just found that page :) thanks

examples/mne_util.py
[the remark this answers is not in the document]
jadrew43: @drammock There are several variables defined after the `__init__` but still within the class, i.e. line 90 above. Can I still include them in the class's note section?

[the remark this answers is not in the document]
drammock: You can add more bullets to the notes section, just copy the example I gave for `G` (the dash at the start of the line will make it a bulleted list, so just add more lines like that).  I'm actually not 100% sure whether the best place is a big bulleted list in `Notes` of the class docstring, or separate entries in docstrings for each property.  It shouldn't be too much work to switch it later, so for now just put everything in the class docstring `Notes` section.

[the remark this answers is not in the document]
jadrew43: Ok cool thanks! What is the standard method for building the citations? I have used Zotero for papers in the past, I'm wondering if you all have a method you rely on. I'll try to get `references.bib` uploaded tonight so you can see what the Zotero output looks like.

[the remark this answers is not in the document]
drammock: Exporting a bibtex entry from Zotero is usually fine, though there may be extraneous fields you'll want to delete (like file path to your pdf copy). See how they look in `doc/references.bib` in the MNE-Python repo

examples/mne_util.py
```diff
@@ -0,0 +1,294 @@
...
+    info = epochs.info.copy()
+    data = epochs.get_data().copy()
+
+    data[:,idx_eeg,:] *= eeg_scale
+    data[:,idx_mag,:] *= mag_scale
+    data[:,idx_grad,:] *= grad_scale
+
+    epochs = mne.EpochsArray(data, info)
+
+    return G, GL, Q, epochs
```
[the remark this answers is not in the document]
jadrew43: With `scales = 1` the `scale_sensor_data()` is not necessary. Should we leave it as part of LDS in case users want to change the scales? 

Also, I'm not following how the steps in `scale_sensor_data()` is the same as the whitening functions you've mentioned. Can you checkout my previous push with commit `does scale_sensor_data() == whitening?` for lines 64-72 in `state_space_connectivity.py()` to see if I'm following you correctly?

examples/state_space_connectivity.py
jadrew43: The matrices that are returned from the above two functions are used as part of `subject_data` to add subject's data to the model. If we are to have the function `model.add_subject` here then we will also need these functions to return these variables. I am open to discussing other methods of going about this.

jadrew43:
`examples/state_space_connectivity.py` is functioning properly on my machine with the output depicted below. I think for a single subject, these results look good. Along the diagonal, values are close to 1, as expected for a computation similar to an autocorrelation. For the condition `auditory/left` there seems to be a connection from `Aud-lh` to `Aud-rh` as seen by the non-zero values in graph [0,1]. I expect measurements to be less noisy when running for a large number of subjects. Please run CI checks.

image

jadrew43:
I am at a research conference for the next 7 days. When I return I have the following to-do:
- work to replace `scale_sensor_data` with `mne.make_ad_hoc_cov()` and `compute_whitener()`
- test for different conditions; see how results change
- use different dataset with multiple subjects for sanity check
Looking forward to your feedback!

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,87 @@
...
+
+## read forward solution, remove bad channels
+fwd_fname = sample_folder / 'sample_audvis-meg-eeg-oct-6-fwd.fif' 
+fwd = mne.read_forward_solution(fwd_fname,exclude=raw.info['bads'])
+fwd = mne.convert_forward_solution(fwd, force_fixed=True)
+
+## read in covariance OR compute noise covariance? noise_cov drops bad chs
+cov_fname = sample_folder / 'sample_audvis-cov.fif'
+cov = mne.read_cov(cov_fname) #has all 366 channels; drop 2?
+noise_cov = mne.compute_covariance(epochs, tmax=0)
```
larsoner, line 48: You shouldn't need to worry about dropping 2 here (presumably due to `bads`?), it should be done internally in `model.add_subject` or `model.fit` when it actually uses the cov

examples/megssm/mne_util.py
```diff
@@ -0,0 +1,290 @@
+""" MNE-Python utility functions for preprocessing data and constructing
```
larsoner, line 1: Move all of these files from `examples` into `mne_connectivity/state_space/*.py` for example

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,87 @@
...
+
+## read labels for analysis
+label_names = ['AUD-lh', 'AUD-rh', 'Vis-lh', 'Vis-rh']
+labels = [mne.read_label(sample_folder / 'labels' / f'{label}.label')
+          for label in label_names]
+
+## initiate model
+num_rois = len(labels)
+timepts = len(epochs.times)
+model = LDS(num_rois, timepts, lam0=0, lam1=100)  # only needs the forward, labels, and noise_cov to be initialized
```
larsoner, line 58: I don't think it should need to take `num_rois` or `timepts`. It can get this information from the first `add_subject` call, and then after the first `add_subject` any subsequent `add_subject` call should check to make sure the same time points and ROIs are used

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,87 @@
...
+
+model.add_subject('sample', subjects_dir, epochs, labels, fwd, noise_cov)
+#when to use compute_cov vs read_cov?
+
+model.fit(niter=100, verbose=1)
+At = model.A
+assert At.shape == (timepts, num_rois, num_rois)
+
+plt.rcParams.update(
```
larsoner, line 67: Better to use `with plt.rc_context` here so the global state is not changed

https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.rc_context.html

examples/state_space_connectivity.py
```diff
@@ -0,0 +1,87 @@
...
+## initiate model
+num_rois = len(labels)
+timepts = len(epochs.times)
+model = LDS(num_rois, timepts, lam0=0, lam1=100)  # only needs the forward, labels, and noise_cov to be initialized
+
+model.add_subject('sample', subjects_dir, epochs, labels, fwd, noise_cov)
+#when to use compute_cov vs read_cov?
+
+model.fit(niter=100, verbose=1)
+At = model.A
```
larsoner, line 64: We should probably call the attribute `model.At` if it's really time-varying `A`

examples/megssm/models.py
```diff
@@ -0,0 +1,872 @@
...
+        scales = {'eeg_scale' : 1, 'mag_scale' : 1, 'grad_scale' : 1} #scale=1 has no effect
+        fwd_src_snsr, fwd_roi_snsr, snsr_cov, epochs = \
+                _scale_sensor_data(epochs, fwd, cov, roi_to_src, **scales)
+                
+        # which_roi = roi_to_src.which_roi # array of len(sources); val = ROI of source
+        # data = epochs._data
+        # data = np.swapaxes(data,-1,-2)
+        # subjectdata = [(data, fwd_roi_snsr, fwd_src_snsr, snsr_cov, which_roi)]
+        
+        sdata = run_pca_on_subject(subject,epochs, fwd, cov, labels)
```
larsoner, line 142: ... specifically I guess this `run_pca_on_subject` function should take into account channel mismatches between `cov`, `fwd`, and `epochs`

examples/megssm/models.py
```diff
@@ -0,0 +1,872 @@
...
+        # new data was added)
+        self._mus_smooth_lst = None
+        self._sigmas_smooth_lst = None
+        self._sigmas_tnt_smooth_lst = None
+        self._loglik = None
+        self._B4 = [None] * self._nsubjects
+
+    # TODO: figure out how to initialize smoothed parameters so this doesn't
+    #       break, e.g. if em_objective is called before em for some reason
+    def em_objective(self):
```
larsoner, line 175: This should be private (prefixed by `_`) to make it clear users should never touch it

examples/megssm/models.py
```diff
@@ -0,0 +1,872 @@
...
+
+        if verbose > 0:
+            print()
+            print("elapsed, function:", datetime.now() - fxn_start)
+            print("=" * 34)
+            print()
+
+        return objvals, converged, mus_smooth_lst, self._loglik, St_lst
+
+    def e_step(self, verbose=0):
```
larsoner, line 482: Private

examples/megssm/models.py
```diff
@@ -0,0 +1,872 @@
...
+            mus_smooth_outer_l1 = einsum2('rti,rtj->rtij',
+                                          mus_smooth[:,1:,:num_roi],
+                                          mus_smooth[:,:-1,:])
+            self._B2 += w_s*np.sum(sigmas_tnt_smooth[:,:,:num_roi,:] +
+                                   mus_smooth_outer_l1, axis=0)
+
+        if verbose > 0:
+            print("\n  done")
+
+    def m_step(self, update_A=True, update_roi_cov=True, update_roi_cov_0=True,
```
larsoner, line 547: etc.

examples/megssm/models.py
```diff
@@ -0,0 +1,872 @@
...
+    @property
+    def roi_cov_0(self):
+        return self._roi_cov_0
+
+    @roi_cov_0.setter
+    def roi_cov_0(self, roi_cov_0):
+        self._roi_cov_0 = roi_cov_0
+
+    @property
+    def log_sigsq_lst(self):
```
larsoner, line 843: All of this stuff looks like it should be private to me. That also means we don't need a `@property` or a `@whatever.setter` for each of them. We just use private attributes directly

examples/megssm/mne_util.py
```diff
@@ -0,0 +1,290 @@
...
+    sel_eeg = pick_types(fwd['info'], meg=False, eeg=True, ref_meg=False)
+    sel_mag = pick_types(fwd['info'], meg='mag', eeg=False, ref_meg=False)
+    sel_grad = pick_types(fwd['info'], meg='grad', eeg=False, ref_meg=False)
+    idx_eeg = [ch_names.index(ch_names[c]) for c in sel_eeg]
+    idx_mag = [ch_names.index(ch_names[c]) for c in sel_mag]
+    idx_grad = [ch_names.index(ch_names[c]) for c in sel_grad]
+
+    # retrieve forward and sensor covariance
+    fwd_src_snsr = fwd['sol']['data'].copy()
+    snsr_cov = cov.data.copy()
```
larsoner, line 196: ... probably here you do
```
epochs = epochs.copy().pick('data', exclude='bads')
snsr_cov = pick_channels_cov(cov, epochs.ch_names, ordered=True).data
fwd = convert_forward_solution(fwd, force_fixed=True)
fwd_src_snsr = pick_channels_forward(fwd, epochs.ch_names, ordered=True)['sol']['data']
data = epochs.get_data()
info = epochs.info
del cov, fwd, epochs
```
now you are guaranteed that `data`, `fwd_src_snsr`, and `snsr_cov `all have the same set of channels in the same order. Then to rescale them all you can just do:
```
rescale_cov = make_ad_hoc_cov(epochs.info)
scaler = compute_whitener(info, rescale_cov)
del rescale_cov
fwd_src_snsr = scaler @ fwd_src_snsr
snsr_cov = scaler @ snsr_cov @ scaler.T
data = scaler @ epochs.get_data()  # @ nicely knows to operate over the last two dims of (epochs, chs, time)
```
or something similar. Basically you let the MNE `pick_channels_*` make sure all channels are present and ordered properly, then make use of MNE functions to do the rescaling.

examples/state_space_connectivity.py
jadrew43: The initialization of the model uses these variables to set up the default values for several model parameters. See `models.py` line 97. Would it make sense to move those initializations to `add_subject()`? I think it makes more sense to leave them in the initialization of the model.

examples/state_space_connectivity.py
jadrew43: `At` is already frequently used in `models.py` when updating `A`. I thought about using something like `connectivity` or `dyn_conn` but I think it would make the code really messy and hard to read. We can talk about it more in our meeting?

examples/megssm/models.py
jadrew43: So I should be able to remove all of these `@property` and `@some.setter` lines right? Does that mean there needs to be additional checks on variable format somewhere? I don't think any of these have any checks as they are written here. I'd also like to talk about which vars and functions to make private. In my mind, all of the vars and functions in `models.py` should be private.

examples/state_space_connectivity.py
adam2392: Perhaps At_? To have fitted attributes have an underscore at the end. In line with scikit learn convention. 

Or A_t_?

examples/megssm/mne_util.py
jadrew43: Previously you said to play around with these functions to see if it would produce the same outputs as the `scale_sensor_data` function. The original function was using a scale of 1 for each of the channel types so nothing was changed. This scaling does alter the data, primarily in a significant increase in the number of principal components (PCA run immediately after scaling function). Also, the output of the model is much noisier using `@ scaler`, I think as expected with the significant increase in PCs. Let's talk to make sure this is working as expected.

jadrew43: Output using mne functions for scaling seems much noisier.
image

jadrew43: Compared to no scaling (or previous function with scale of 1 for each channel type).
image

jadrew43:
I am currently working to incorporate an old dataset to see if these scripts produce the expected results.

jadrew43:
Hey @adam2392 one of the CI errors is due to autograd: `ModuleNotFoundError: No module named 'autograd'`. I'd really like to move to the next step of the proposal and work to integrate jax (to replace autograd) later on in order to keep pace with the summer milestones. Is it alright if I install autograd in the dependencies for mnedev (assuming that will fix the error) for now and work to integrate jax later on?

adam2392:
Yeah sure! You can include it as an optional Dev dependency. Do you know where to add? Please also add a comment if you can that we replace it later on. 

Perhaps we can start another sep issue to track migration to jax later on?

jadrew43:
> Yeah sure! You can include it as an optional Dev dependency. Do you know where to add? Please also add a comment if you can that we replace it later on.
> 
> Perhaps we can start another sep issue to track migration to jax later on?

@adam2392 Cool! I do not where to add autograd as an optional Dev dependency. Should it be installed in mnedev?

Where should I leave the comment that it can be replaced with jax later on? Just start a new issue on Github with the suggestion?

adam2392:
> > Yeah sure! You can include it as an optional Dev dependency. Do you know where to add? Please also add a comment if you can that we replace it later on.
> > Perhaps we can start another sep issue to track migration to jax later on?
> 
> @adam2392 Cool! I do not where to add autograd as an optional Dev dependency. Should it be installed in mnedev?

You can add it here: https://github.com/mne-tools/mne-connectivity/blob/main/requirements_testing.txt
and then just add a comment e.g.

```
...
# TODO: we will replace this with Jax
autograd
```

Then when you import autograd, you should maybe import it within the functions you want to use and not at the top of the Python file. That way there is no error when someone tries to run `import mne_connectivity` if they don't have autograd/jax installed. For example, this [function](https://github.com/mne-tools/mne-python/blob/4e5ca1bf86429d976d86223480bd091d4b3888e2/mne/gui/__init__.py#L203-L237) inside MNE-Python needs `pyqt`, but imports it within the function so mne still works if the user only has `numpy/scipy` installed.

> Where should I leave the comment that it can be replaced with jax later on? Just start a new issue on Github with the suggestion?

Yeah I'll create the GH issue.

jadrew43:
> > > Yeah sure! You can include it as an optional Dev dependency. Do you know where to add? Please also add a comment if you can that we replace it later on.
> > > Perhaps we can start another sep issue to track migration to jax later on?
> > 
> > 
> > @adam2392 Cool! I do not where to add autograd as an optional Dev dependency. Should it be installed in mnedev?
> 
> You can add it here: https://github.com/mne-tools/mne-connectivity/blob/main/requirements_testing.txt and then just add a comment e.g.
> 
> ```
> ...
> # TODO: we will replace this with Jax
> autograd
> ```
> 
> Then when you import autograd, you should maybe import it within the functions you want to use and not at the top of the Python file. That way there is no error when someone tries to run `import mne_connectivity` if they don't have autograd/jax installed. For example, this [function](https://github.com/mne-tools/mne-python/blob/4e5ca1bf86429d976d86223480bd091d4b3888e2/mne/gui/__init__.py#L203-L237) inside MNE-Python needs `pyqt`, but imports it within the function so mne still works if the user only has `numpy/scipy` installed.
> 
> > Where should I leave the comment that it can be replaced with jax later on? Just start a new issue on Github with the suggestion?
> 
> Yeah I'll create the GH issue.

Ok I think that's complete. Thanks!

examples/megssm/mne_util.py
larsoner: > The original function was using a scale of 1 for each of the channel types so nothing was changed.

This is the equivalent of using `mne.make_ad_hoc_cov(..., std=1.)`. So if you want this mode you can get it easily with a very small change. Can you verify this looks the same as the old code?

> This scaling does alter the data, primarily in a significant increase in the number of principal components (PCA run immediately after scaling function).

This is to be expected. If you do no scaling and have MEG *and* EEG data, your EEG data will be ~6 orders of magnitude larger or so, so you will end up only looking at EEG data.

> Also, the output of the model is much noisier using @ scaler, I think as expected with the significant increase in PCs. Let's talk to make sure this is working as expected.

What's not clear to me: is it also noisier (and noisier in the same way) when using the *old* code path but *enabling* scaling?

In other words: is this a problem with the new way of scaling, or is this a problem that has always existed with scaling the data, regardless of scaling method?

We can chat about this today

jadrew43:
Update: I got the code to work for an old dataset, however the results are not what I expected. I am currently working to get the sample data (much smaller than the old dataset) to work on my original model. This output will be my new ground truth and I can work iteratively to make sure each edit I make to the original model to conform it the MNE-Python API standards produces the same output. Lesson learned - have a simple ground truth to work with from the beginning :)

jadrew43:
Here is the output of my original model using the sample dataset for the auditory/left condition for the 12 ROIs from a different dataset. My next step is to get this same output in the `mnedev` environment. Then I will work to use only the 4 ROIs commonly used with the sample dataset. Then piece by piece I will recreate the API.
image

larsoner:
Excellent!

A good next step is to make sure all random seeds can be set such that if you run this again you get the exact same output (to numerical precision at least). Then you can save the At result to disk now and compare to it each time you replace some piece of code

jadrew43:
I have changed the PCA method from being based on the rank of the matrix to a method based on explaining 99% of the variance. This method allows the fitting of the model to run much faster as it produces 147 principal components vs 360 components produced from the rank method. The model output is noticeably but not extremely different. My next step is to perform the processing steps using the 4 labels provided in the sample dataset, which should reduce processing time even further.
image

jadrew43:
Processing completed with 4 labels from `sample`.
image

jadrew43:
Bootstrapping of epochs, and PCA of `epochs.get_data()` and forward matrices completed in API. Model fitting completed in command line model. Output from API (LDS) compared to command line model (MEGLDS) are not identical but extremely similar. 
image
image

larsoner:
Nice!

It would be good to know what the differences are that make them not identical, but really if this version of the API works on our UW data as well (maybe even just for one subject?) then I'd say you could use this as the "ground truth" for correctness of additional changes!

jadrew43:
@larsoner Can you look at `megssm/mne_util.py` L114. Am I using the scaler correctly? Because the results do not agree with the original `_scale_sensor_data` (L140). Thanks. 

I'll get to the CI checks first thing Monday!

examples/megssm/mne_util.py
```diff
@@ -0,0 +1,290 @@
...
+
+    @property
+    def fwd_src_roi(self):
+        return self._fwd_src_roi
+
+    @fwd_src_roi.setter
+    def fwd_src_roi(self, val):
+        self._fwd_src_roi = val
+
+    @property
```
larsoner, line 114: @jadrew43 this is line 114 and it's not related to scaling. There is some stuff below that is. Can you confirm by commenting on the correct lines/lines in the "Files" tab of this PR? This not being the correct line makes me wonder if you forgot a push, and I don't want to look prematurely.

The best way to help me check would be to make it very easy to run and check the results. For example, add a commented-out (to make CIs happy) `# np.testing.assert_allclose(data, data_mne, rtol=..., atol=...)` statement that fails (when you uncomment it) where you scale `data` the old way and `data_mne` the new way. Then I need a minimal script (hopefully runs in just a few seconds, and on `sample` data for example) I can run on this branch that will fail when I uncomment the `assert_allclose` line

jadrew43: The updated files are in `state_space/`, I haven't done anything in `examples/` in weeks so I went ahead and deleted the files I had in there. Running `state_space_connectivity.py` takes about 90 seconds, adding `_mne_scale_sensor_data()` adds about 60 seconds. Happy to chat about building tests that take much less time. The commented testing line is L228 in `mne_util.py`.

larsoner: Hmm, I don't think this is quite a minimal reproducible bit of code.

When I tried to run the code, it failed because I needed `autograd`. I installed that, then it complained that I needed `autograd_linalg`. I can't find this on `pip` or `conda`, so I tried swapping in `from scipy.linalg import solve_triangular`.

Next I tried to run the script and got:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/larsoner/mne_data/MNE-sample-data/MEG/sample/labels/AUD-lh.label'
```
This makes sense, usually labels are in the SUBJECTS_DIR. But even then there aren't auditory labels in `sample`. So let's change it to something that should work I think:
```
regexp = '^(G_temp_sup-Plan_tempo.*|Pole_occipital)'
labels = mne.read_labels_from_annot(
    'sample', 'aparc.a2009s', regexp=regexp, subjects_dir=subjects_dir)
label_names = [label.name for label in labels]
assert len(label_names) == 4
```
Then I got a bit farther, until I hit:
```
AttributeError: python: undefined symbol: mkl_get_max_threads
```
So I then disabled all the thread-setting stuff, and then get to:
```
  File "/home/larsoner/python/scipy/scipy/linalg/_basic.py", line 336, in solve_triangular
    raise ValueError('expected square matrix')
ValueError: expected square matrix
```
So my swap of `scipy.linalg` is not correct.

I don't think I can proceed until I can install `autograd_linalg` somehow, how do you recommend doing this?

jadrew43: Try this:
pip install --src deps -e git+ssh://git@github.com/nfoti/autograd_linalg.git@master#egg=autograd_linalg

larsoner: I pushed a commit for a `solve_triangular` that allows the code to run on my machine. I can look into the issue now!

adam2392:
Hi @jadrew43 and @larsoner any help needed to review code / look at prelim results here? Feel free to lmk where I can help!

larsoner:
IIRC code is still WIP / needs to be systematically converted to MNE conventions, but some bugs have been found along the way which is good!

drammock:
@jadrew43 is there any interest in coming back to this, or should we close (until someone else has time to pick it up)?

Issue #111, opened 2022-11-11
spectral_connectivity_epochs ignores verbose=False


sam-s:
#### Describe the bug

```
mne_connectivity.spectral_connectivity_epochs(..., verbose=False, n_jobs=-1)
```

prints 

```
Adding metadata with 3 columns
Connectivity computation...
only using indices for lower-triangular matrix
    computing connectivity for 496 connections
    using t=-1.000s..4.000s for estimation (2501 points)
    computing connectivity for the bands:
     band 1: 1.2Hz..4.0Hz (15 points)
     band 2: 4.2Hz..7.0Hz (15 points)
     band 3: 8.2Hz..12.0Hz (20 points)
     band 4: 12.2Hz..30.0Hz (90 points)
     band 5: 30.2Hz..55.0Hz (125 points)
   ....
```

#### Additional information
```
Platform:         Linux-6.0.6-76060006-generic-x86_64-with-glibc2.35
Python:           3.10.6 (main, Nov  2 2022, 18:53:38) [GCC 11.3.0]
Executable:       /usr/bin/python3
CPU:              x86_64: 8 cores
Memory:           15.3 GB

mne:              1.2.1
numpy:            1.23.4 {OpenBLAS 0.3.20 with 8 threads}
scipy:            1.9.3
matplotlib:       3.6.2 {backend=module://matplotlib_inline.backend_inline}

sklearn:          1.1.3
numba:            Not found
nibabel:          Not found
nilearn:          Not found
dipy:             Not found
openmeeg:         Not found
cupy:             Not found
pandas:           1.5.1
pyvista:          Not found
pyvistaqt:        Not found
ipyvtklink:       Not found
vtk:              Not found
qtpy:             2.3.0 {PyQt5=5.15.3}
ipympl:           Not found
pyqtgraph:        Not found
pooch:            v1.6.0

mne_bids:         Not found
mne_nirs:         Not found
mne_features:     Not found
mne_qt_browser:   Not found
mne_connectivity: 0.4.0
mne_icalabel:     Not found
```

sam-s:
Possibly related: https://github.com/mne-tools/mne-python/issues/11321

sam-s:
workaround: `mne.utils.logger.setLevel(30)`

adam2392:
Hmm.. where is this message "Adding metadata with 3 columns" coming from? I do a ctrl+f in both mne-connectivity and mne-python and do not see it.

sam-s:
[...]

PR #112, opened 2022-11-17
add missing @verbose annotations, fix bug#111

sam-s:
PR Description
--------------

Fix https://github.com/mne-tools/mne-connectivity/issues/111

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_connectivity/spectral/epochs.py
+++ b/mne_connectivity/spectral/epochs.py
@@ -16,7 +16,8 @@ from mne.time_frequency.multitaper import (_csd_from_mt,
                                            _psd_from_mt_adaptive)
 from mne.time_frequency.tfr import cwt, morlet
 from mne.time_frequency.multitaper import _compute_mt_params
-from mne.utils import (_arange_div, _check_option, logger, warn, _time_mask, verbose)
+from mne.utils import (_arange_div, _check_option, logger, warn, _time_mask,
+                       verbose)
 
 from ..base import (SpectralConnectivity, SpectroTemporalConnectivity)
 from ..utils import fill_doc, check_indices
@@ -831,8 +832,8 @@ def spectral_connectivity_epochs(data, names=None, method='coh', indices=None,
         Use adaptive weights to combine the tapered spectra into PSD.
         Only used in 'multitaper' mode.
     mt_low_bias : bool
-        Only use tapers with more than 90 percent spectral concentration within bandwidth.
-        Only used in 'multitaper' mode.
+        Only use tapers with more than 90 percent spectral concentration
+        within bandwidth. Only used in 'multitaper' mode.
     cwt_freqs : array
         Array of frequencies of interest. Only used in 'cwt_morlet' mode.
     cwt_n_cycles : float | array of float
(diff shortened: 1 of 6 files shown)


mne_connectivity/spectral/epochs.py
```diff
@@ -760,6 +760,7 @@ def _check_estimators(method, mode):
     return con_method_types, n_methods, accumulate_psd, n_comp_args
 
 
+@verbose
 @fill_doc
```
drammock, line 764: for all of these, `@verbose` can simply replace `@fill_doc` (`@verbose` does everything `@fill_doc` does plus a few other things)

sam-s: There are some places where both `@verbose` and `@fill_doc` are present - should I remove the latter there too?

drammock: actually now that I look, I realize that MNE-Connectivity has it's own `@fill_doc` definition, but it uses MNE-Python's `@verbose`.  Off the top of my head I don't know how that will interact, so there's a chance you might actually need both.  I recommend only changing these 3 here, then checking the results before changing other occurrences elsewhere in the codebase.

drammock: (also you will need to import verbose in those files where you're adding it)

larsoner: If you put fill_doc below verbose, I think it should work
```
@verbose
@fill_doc
def whatever(....)
````

Comment (sam-s) [reply mne_connectivity/spectral/epochs.py:764]: yes, this is what is done elsewhere

**Comment by sam-s** [comment]
None of the failures appear to be connected to the PR.

**Review by adam2392 — COMMENTED** [review]
Feel free to add yourself as a contributor via a `whats_new.rst` @sam-s

**Review by adam2392 — APPROVED** [review]
Hopefully CI passes and this is good to go.

File: mne_connectivity/spectral/epochs.py [hunk 2]
```diff
@@ -830,8 +831,8 @@ def spectral_connectivity_epochs(data, names=None, method='coh', indices=None,
         Use adaptive weights to combine the tapered spectra into PSD.
         Only used in 'multitaper' mode.
     mt_low_bias : bool
-        Only use tapers with more than 90%% spectral concentration within
-        bandwidth. Only used in 'multitaper' mode.
+        Only use tapers with more than 90 percent spectral concentration within bandwidth.
```
adam2392, line 834: @larsoner @sam-s  not sure why, but the '%%' throws an error using verbose(?). Also if I replace it w/ '\%', it throws an error as well. Not worth the trouble to figure out I guess.

larsoner: You probably need `%%%%`. This is because both `@fill_doc` and `@verbose` will try to complete these, so the first time through you need `%%%%` so this reduces to `%%`, which on the second `%`-operation reduces to just `%`. Or to put it in code:
```
>>> '%%%%' % () % ()
'%'
```

PR #104, opened 2022-08-31
[MRG] [BUG] [ENH] [WIP] Bug fixes and enhancements for time-resolved spectral connectivity estimation

ruuskas:
PR Description
--------------

This PR closes #90, closes #84, and closes #17. Issues discussed in #70 and #73 are addressed. Please see the commit messages for description of the changes. Unfortunately, many fixes and enhancements are bundled together, as it's impossible to separate them without breaking functionality.

In summary,
- API is streamlined with `mne_connectivity.spectral_connectivity_epochs`.
- The bug mentioned in #84 is fixed and connectivity is now averaged over tapers, instead of the power spectra.
- The issue with too high PLV values mentioned in #73 is fixed, time-resolved PLV is now correctly averaged over time points before taking the absolute value.
- An array of connectivity methods is now supported and an array of corresponding connectivity results is returned.
- The `block_size` parameter now corresponds to actual block size instead of number of blocks.
- wPLI and PLI connectivity metrics are added.
- Connectivity can be averaged over Epochs and frequencies
- Frequencies for multitapering are obtained automatically similarly to `spectral_connectivity_epochs`.

The following ZIP archive contains a comparison between the current release version, the version of this PR, and the HyPyP implementation of spectral connectivity (`mne_connectivity_testing_random_data.html`). This shows that the issue with too high connectivity scores with random data is now fixed. The archive also contains a comparison with HyPyP using MNE-Python example data.

[examples.zip](https://github.com/mne-tools/mne-connectivity/files/9461059/examples.zip)

I marked this as WIP as I still have some concerns. Namely,
- What is the purpose of spectral smoothing in connectivity estimation, is it necessary and is there some good reference material?
- Are the computed connectivity values correct, how to verify this?
- The implementation is extremely slow compared to HyPyP, even with parallel computation enabled. This is due to the way connectivity scores are computed between individual signals in a for-loop. I'm considering vectorizing the computation and using `numpy.einsum` as is done in [HyPyP](https://github.com/ppsp-team/HyPyP/blob/a7f78ef9886ca0621b0d2b5d5650eb1d8c903fc8/hypyp/analyses.py#398). With the current implementation, computing source space connectivity scores for the example dataset takes about 20 minutes per connectivity metric when the aparc_sub parcellation (~400 parcels) is used, whereas HyPyP takes less than two minutes for the same task on my local hardware (please see the attached example for details).
- The regression test fails due to assertion error. However, this is explained by the (apparently) erroneous code in the original.
- Should there be more logging?

[...]

PR #115, opened 2022-11-25
[ENH] Multiple improvements to spectral_connectivity_time: ciPLV, and efficient computation of multiple metrics 

ruuskas:
PR Description
--------------

This PR addresses improvements mentioned in #104. I'm reposting this with the Git log now cleaned up.

- Add ciPLV metric.
- Compute all connectivity measures at once, avoiding the repeated computation of the cross-spectral density.
- Reinstitute the option to specify frequencies of interest also when using the multitaper method. The reasoning is that users should have control over the sparsity of the frequency grid.
- Add the option to use parts of each epoch as padding. This allows for the mitigation of edge effects in the time-frequency decomposition. The edges are removed after time-frequency decomposition, but before connectivity computation.
- Improve the function description to better describe the objective.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_connectivity/spectral/tests/test_spectral.py
+++ b/mne_connectivity/spectral/tests/test_spectral.py
@@ -504,11 +504,11 @@ def test_spectral_connectivity_time_phaselocked(method, mode, data_option):
     # hypothesized "connection"
     freq_band_low_limit = (8.)
     freq_band_high_limit = (13.)
-    cwt_freqs = np.arange(freq_band_low_limit, freq_band_high_limit + 1)
+    freqs = np.arange(freq_band_low_limit, freq_band_high_limit + 1) if mode == 'cwt_morlet' else None
     con = spectral_connectivity_time(data, method=method, mode=mode,
                                      sfreq=sfreq, fmin=freq_band_low_limit,
                                      fmax=freq_band_high_limit,
-                                     cwt_freqs=cwt_freqs, n_jobs=1,
+                                     freqs=freqs, n_jobs=1,
                                      faverage=True, average=True, sm_times=0)
     assert con.shape == (n_channels ** 2, len(con.freqs))
     con_matrix = con.get_data('dense')[..., 0]
@@ -589,11 +589,11 @@ def test_spectral_connectivity_time_resolved(method, mode):
     data = EpochsArray(data, info)
 
     # define some frequencies for cwt
-    freqs = np.arange(3, 20.5, 1)
+    freqs = np.arange(3, 20.5, 1) if mode == 'cwt_morlet' else None
 
     # run connectivity estimation
     con = spectral_connectivity_time(
-        data, sfreq=sfreq, cwt_freqs=freqs, method=method, mode=mode,
+        data, sfreq=sfreq, freqs=freqs, method=method, mode=mode,
         n_cycles=5)
     assert con.shape == (n_epochs, n_signals ** 2, len(con.freqs))
     assert con.get_data(output='dense').shape == \
--- a/mne_connectivity/spectral/tests/test_spectral.py
+++ b/mne_connectivity/spectral/tests/test_spectral.py
@@ -2,14 +2,8 @@ import numpy as np
 from numpy.testing import (assert_allclose, assert_array_almost_equal,
                            assert_array_less)
 import pytest
-import warnings
-
-import mne
-from mne import (EpochsArray, SourceEstimate, create_info,
-                 make_fixed_length_epochs)
+from mne import (EpochsArray, SourceEstimate, create_info)
 from mne.filter import filter_data
-from mne.utils import _resource_path
-from mne_bids import BIDSPath, read_raw_bids
 
 from mne_connectivity import (
     SpectralConnectivity, spectral_connectivity_epochs,
@@ -478,7 +472,101 @@ def test_epochs_tmin_tmax(kind):
     assert len(w) == 1  # just one even though there were multiple epochs
 
 
-@pytest.mark.parametrize('method', ['coh', 'plv'])
+@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli'])
+@pytest.mark.parametrize(
+    'mode', ['cwt_morlet', 'multitaper'])
+@pytest.mark.parametrize('data_option', ['sync', 'random'])
+def test_spectral_connectivity_time_phaselocked(method, mode, data_option):
+    """Test time-resolved spectral connectivity with simulated phase-locked
+    data."""
+    rng = np.random.default_rng(0)
+    n_epochs = 5
+    n_channels = 3
+    n_times = 1000
+    sfreq = 250
+    data = np.zeros((n_epochs, n_channels, n_times))
+    if data_option == 'random':
+        # Data is random, there should be no consistent phase differences.
+        data = rng.random((n_epochs, n_channels, n_times))
+    if data_option == 'sync':
+        # Data consists of phase-locked 10Hz sine waves with constant phase
+        # difference within each epoch.
+        wave_freq = 10
+        epoch_length = n_times / sfreq
+        for i in range(n_epochs):
+            for c in range(n_channels):
+                phase = rng.random() * 10
+                x = np.linspace(-wave_freq * epoch_length * np.pi + phase,
+                                wave_freq * epoch_length * np.pi + phase,
+                                n_times)
+                data[i, c] = np.squeeze(np.sin(x))
+    # the frequency band should contain the frequency at which there is a
+    # hypothesized "connection"
+    freq_band_low_limit = (8.)
+    freq_band_high_limit = (13.)
+    cwt_freqs = np.arange(freq_band_low_limit, freq_band_high_limit + 1)
+    con = spectral_connectivity_time(data, method=method, mode=mode,
+                                     sfreq=sfreq, fmin=freq_band_low_limit,
+                                     fmax=freq_band_high_limit,
+                                     cwt_freqs=cwt_freqs, n_jobs=1,
+                                     faverage=True, average=True, sm_times=0)
+    assert con.shape == (n_channels ** 2, len(con.freqs))
+    con_matrix = con.get_data('dense')[..., 0]
+    if data_option == 'sync':
+        # signals are perfectly phase-locked, connectivity matrix should be
+        # a lower triangular matrix of ones
+        assert np.allclose(con_matrix,
+                           np.tril(np.ones(con_matrix.shape),
+                                   k=-1),
+                           atol=0.01)
+    if data_option == 'random':
+        # signals are random, all connectivity values should be small
+        # 0.5 is picked rather arbitrarily such that the obsolete wrong
+        # implementation fails
+        assert np.all(con_matrix) <= 0.5
+
+
+@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli'])
+@pytest.mark.parametrize(
+    'cwt_freqs', [[8., 10.], [8, 10], 10., 10])
+def test_spectral_connectivity_time_cwt_freqs(method, cwt_freqs):
+    """Test time-resolved spectral connectivity with int and float values for
+    cwt_freqs."""
+    rng = np.random.default_rng(0)
+    n_epochs = 5
+    n_channels = 3
+    n_times = 1000
+    sfreq = 250
+    data = np.zeros((n_epochs, n_channels, n_times))
+
+    # Data consists of phase-locked 10Hz sine waves with constant phase
+    # difference within each epoch.
+    wave_freq = 10
+    epoch_length = n_times / sfreq
+    for i in range(n_epochs):
+        for c in range(n_channels):
+            phase = rng.random() * 10
+            x = np.linspace(-wave_freq * epoch_length * np.pi + phase,
+                            wave_freq * epoch_length * np.pi + phase,
+                            n_times)
+            data[i, c] = np.squeeze(np.sin(x))
+    # the frequency band should contain the frequency at which there is a
+    # hypothesized "connection"
+    con = spectral_connectivity_time(data, method=method, mode='cwt_morlet',
+                                     sfreq=sfreq, fmin=np.min(cwt_freqs),
+                                     fmax=np.max(cwt_freqs),
+                                     cwt_freqs=cwt_freqs, n_jobs=1,
+                                     faverage=True, average=True, sm_times=0)
+    assert con.shape == (n_channels ** 2, len(con.freqs))
+    con_matrix = con.get_data('dense')[..., 0]
+
+    # signals are perfectly phase-locked, connectivity matrix should be
+    # a lower triangular matrix of ones
+    assert np.allclose(con_matrix, np.tril(np.ones(con_matrix.shape), k=-1),
+                       atol=0.01)
+
+
+@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli'])
 @pytest.mark.parametrize(
     'mode', ['cwt_morlet', 'multitaper'])
 def test_spectral_connectivity_time_resolved(method, mode):
@@ -486,7 +574,7 @@ def test_spectral_connectivity_time_resolved(method, mode):
     sfreq = 50.
     n_signals = 3
     n_epochs = 2
-    n_times = 256
+    n_times = 1000
     trans_bandwidth = 2.
     tmin = 0.
     tmax = (n_times - 1) / sfreq
@@ -502,21 +590,20 @@ def test_spectral_connectivity_time_resolved(method, mode):
 
     # define some frequencies for cwt
     freqs = np.arange(3, 20.5, 1)
-    n_freqs = len(freqs)
 
     # run connectivity estimation
     con = spectral_connectivity_time(
-        data, freqs=freqs, method=method, mode=mode)
-    assert con.shape == (n_epochs, n_signals * 2, n_freqs, n_times)
+        data, sfreq=sfreq, cwt_freqs=freqs, method=method, mode=mode,
+        n_cycles=5)
+    assert con.shape == (n_epochs, n_signals ** 2, len(con.freqs))
     assert con.get_data(output='dense').shape == \
-        (n_epochs, n_signals, n_signals, n_freqs, n_times)
-
-    # average over time
-    conn_data = con.get_data(output='dense').mean(axis=-1)
-    conn_data = conn_data.mean(axis=-1)
+        (n_epochs, n_signals, n_signals, len(con.freqs))
 
     # test the simulated signal
     triu_inds = np.vstack(np.triu_indices(n_signals, k=1)).T
+
+    # average over frequencies
+    conn_data = con.get_data(output='dense').mean(axis=-1)
 
     # the indices at which there is a correlation should be greater
     # then the rest of the components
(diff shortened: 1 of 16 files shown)


ruuskas:
It looks like the CIs are not happy about the docs and the loading the Matplotlib Qt backend fails with Python 3.7.

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -472,7 +472,7 @@ def test_epochs_tmin_tmax(kind):
     assert len(w) == 1  # just one even though there were multiple epochs
 
 
-@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli'])
+@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli', 'ciplv'])
```
adam2392, line 475: Do you mind adding a separate test for these where `freqs` is set and asserting some behavior? I.e. that certain frequency bins are used vs the default done in `multitaper`?

Perhaps a separate test for `test_spectral_connectivity_time_multitaper`?

mne_connectivity/spectral/time.py
```diff
@@ -430,23 +443,25 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
...
             # weights have shape (n_tapers, n_freqs, n_times)
     else:
         raise ValueError("Mode must be 'cwt_morlet' or 'multitaper'.")
 
+    out = np.squeeze(out, axis=0)
+
+    if padding:
+        pad_idx = int(np.floor(padding * sfreq / decim))
+        out = out[..., pad_idx:-pad_idx]
+        weights = weights[..., pad_idx:-pad_idx]
```
adam2392, line 457: Do you mind also adding a test for padding?

mne_connectivity/spectral/time.py
```diff
@@ -457,135 +472,89 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
...
+                out.append(pli)
+
+            if m == 'wpli':
+                con_num = np.abs(s_xy.imag.mean(axis=-1, keepdims=True))
+                con_den = np.mean(np.abs(s_xy.imag), axis=-1, keepdims=True)
+                wpli = con_num / con_den
+                out.append(wpli)
+
+            if m == 'cs':
+                out.append(s_xy)
```
adam2392, line 540: WDYT about keeping the basic pairwise "computation" functions as separate functions, so that way they can be self-documented in their docstring? E.g. keeping all the old functions?

And then to add say a new spectral-connectivity function, one would have to: i) add a new private function e.g. `def _newpli(x, y)` and then ii) update `_parallel_con` to add this option. 

Might be easier to read. But also not 100% necessary.

mne_connectivity/spectral/time.py
```diff
@@ -631,3 +600,25 @@ def _foi_average(conn, foi_idx):
         f_e += 1 if f_s == f_e else f_e
         conn_f[..., n_f, :] = conn[..., f_s:f_e, :].mean(-2)
     return conn_f
+
+
+def _compute_freqs(n_times, sfreq, freqs, mode):
+    from scipy.fft import rfftfreq
+    # get frequencies of interest for the different modes
+    if freqs is not None:
+        if any(freqs > (sfreq / 2.)):
```
adam2392, line 609: I think this is defined already in `spectral.epochs.py`. Can we use that one?

adam2392:
> It looks like the CIs are not happy about the docs and loading the Matplotlib Qt backend fails with Python 3.7.
> 
> 
> 
> I can't check what's wrong with the docs it seems. Building locally, it complains about `'secondary_sidebar_items': ['page-toc']` added [here](https://github.com/mne-tools/mne-connectivity/pull/104/commits/b2dda41850222d6867fc3041c4f8bc018fe4f91a#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R201).

Should be fixed now on main.

mne_connectivity/spectral/tests/test_spectral.py
ruuskas: I agree that a test is necessary, will do.

mne_connectivity/spectral/time.py
ruuskas: Sure

mne_connectivity/spectral/time.py
ruuskas: It's an option for sure. However, it should be done such that we avoid computing the cross-spectral density multiple times. This is slow due to the potentially large size of the arrays according to some profiling I did.

mne_connectivity/spectral/time.py
adam2392: Can you add `freqs` into the existing function and then use it? 

Having one private function makes the code more robust, easier to maintain, and less likely to introduce runtime bugs in the future.

mne_connectivity/spectral/time.py
adam2392: Will take a look to see if this is possible then. 

I wonder if it's possible to keep even the same "spectral-conn" base functions across 'epochs' and 'time' computation. I.e. there are the classes inside `spectral_conn_epochs.py` which also compute certain spectral connectivity. They seem to all operate on a pairwise nature, so perhaps we can even consolidate into one place, so then adding one connectivity function adds it for both 'epochs' and 'time'

mne_connectivity/spectral/tests/test_spectral.py
ruuskas: I realized this is a bit troublesome. Namely, we currently require neither `freqs` nor `fmin`/`fmax`. This leads to the issue of setting sensible defaults for `fmin` and `fmax`. 

Essentially, `fmin` should be set such that the wavelet with the set `n_cycles` doesn't become longer than the signal. However, even with the length of the wavelet equal to that of the signal, the spectral estimate isn't very good and thus we cannot `assert connectivity == 1` for synchronized data.

There's still another problem with `freqs` and `n_cycles` which I noticed while looking into this. Essentially, `n_cycles` should either be a single number or the same length as `freqs`. This could lead to confusion, as the length of 'freqs' is not known by the user unless its set by them. One option would be removing '_compute_freqs' and just requiring 'freqs' like it used to be before I started tampering with the code.

mne_connectivity/spectral/time.py
ruuskas: I'm afraid it wouldn't be practical to use the classes from `spectral_connectivity_epochs`. Those estimate connectivity as mean over epochs with the assumption that the connectivity measure is stationary within an Epoch. An estimator's `accumulate` method is therefore called only once per Epoch. Cross-spectral density is computed in a single time window spanning the whole Epoch (with the exception of `cwt_morlet` mode).

In `spectral_connectivity_time` the pairwise computation involves summing over the time axis and `multitaper` actually means convolution with a DPSS-tapered sinusoid (similar to Gaussian tapered sinusoid for `morlet`. Summing over the time axis using something like the `accumulate` function in `epochs` would necessitate a for-loop over time. There's also the possibility of fetching connectivity over time separately for each epoch, which of course doesn't exist in `epochs`. 

In my opinion, accommodating both within the same classes would probably lead to major refactoring of `spectral_connectivity_epochs` and potentially make the code less readable.

Sorry if the explanation is too basic, I'm mostly writing it down to keep it straight for myself.

mne_connectivity/spectral/time.py
ruuskas: Technically yes, but then we would have to prevent the use of `cwt_freqs` together with `multitaper` in `spectral_connectivity_epochs`. In that case `freqs` must be equal to `rfftfreq(n_times, 1./sfreq)` as that's the shape of the output of `_csd_from_mt` and `_psd_from_mt`.

mne_connectivity/spectral/time.py
ruuskas: WDYT about doing it this way?

```python

def _parallel_con(w, method, kernel, foi_idx, source_idx, target_idx, n_jobs,
                  verbose, total, faverage, weights):
    """Compute spectral connectivity in parallel.

    Input signal w is of shape (n_chans, n_tapers, n_freqs, n_times)."""

    if 'coh' in method:
        # psd
        if weights is not None:
            psd = weights * w
            psd = psd * np.conj(psd)
            psd = psd.real.sum(axis=1)
            psd = psd * 2 / (weights * weights.conj()).real.sum(axis=0)
        else:
            psd = w.real ** 2 + w.imag ** 2
            psd = np.squeeze(psd, axis=1)

        # smooth
        psd = _smooth_spectra(psd, kernel)

    def pairwise_con(w_x, w_y):
        # csd
        if weights is not None:
            s_xy = np.sum(weights * w[w_x] * np.conj(weights * w[w_y]), axis=0)
            s_xy = s_xy * 2 / (weights * np.conj(weights)).real.sum(axis=0)
        else:
            s_xy = w[w_x] * np.conj(w[w_y])
            s_xy = np.squeeze(s_xy, axis=0)
        s_xy = _smooth_spectra(s_xy, kernel)
        out = []
        conn_func = {'plv': _plv, 'ciplv': _ciplv, 'pli': _pli, 'wpli': _wpli,
                     'coh': _coh, 'cs': _cs}
        for m in method:
            if m == 'coh':
                s_xx = psd[w_x]
                s_yy = psd[w_y]
                out.append(conn_func[m](s_xx, s_yy, s_xy))
            else:
                out.append(conn_func[m](s_xy))

        for i, _ in enumerate(out):
            # mean inside frequency sliding window (if needed)
            if isinstance(foi_idx, np.ndarray) and faverage:
                out[i] = _foi_average(out[i], foi_idx)
            # squeeze time dimension
            out[i] = out[i].squeeze(axis=-1)

        return out

    # only show progress if verbosity level is DEBUG
    if verbose != 'DEBUG' and verbose != 'debug' and verbose != 10:
        total = None

    # define the function to compute in parallel
    parallel, p_fun, n_jobs = parallel_func(
        pairwise_con, n_jobs=n_jobs, verbose=verbose, total=total)

    return parallel(p_fun(s, t) for s, t in zip(source_idx, target_idx))


def _plv(s_xy):
    s_xy = s_xy / np.abs(s_xy)
    plv = np.abs(s_xy.mean(axis=-1, keepdims=True))
    return plv


def _ciplv(s_xy):
    s_xy = s_xy / np.abs(s_xy)
    rplv = np.abs(np.mean(np.real(s_xy), axis=-1, keepdims=True))
    iplv = np.abs(np.mean(np.imag(s_xy), axis=-1, keepdims=True))
    ciplv = iplv / (np.sqrt(1 - rplv ** 2))
    return ciplv


def _pli(s_xy):
    pli = np.abs(np.mean(np.sign(np.imag(s_xy)),
                         axis=-1, keepdims=True))
    return pli


def _wpli(s_xy):
    con_num = np.abs(s_xy.imag.mean(axis=-1, keepdims=True))
    con_den = np.mean(np.abs(s_xy.imag), axis=-1, keepdims=True)
    wpli = con_num / con_den
    return wpli


def _coh(s_xx, s_yy, s_xy):
    con_num = np.abs(s_xy.mean(axis=-1, keepdims=True))
    con_den = np.sqrt(s_xx.mean(axis=-1, keepdims=True) *
                      s_yy.mean(axis=-1, keepdims=True))
    coh = con_num / con_den
    return coh


def _cs(s_xy):
    return s_xy.mean(axis=-1, keepdims=True)
```

adam2392: Yeah that looks better and more similar to the epochs file. 

Idr but can we put the pairwise function outside? Nested functions imo are harder to read. But if that breaks the paralleliation via joblib, then nvm.

ruuskas: I'll try if the pairwise function works outside `_parallel_con`. Fully agree that nested functions are ugly.

mne_connectivity/spectral/time.py
adam2392: I think that is fine. So, we would check `if freqs is not None and mode == 'multitaper'` and raise an error.

The function is essentially doing the same thing regardless of which one it's being called from.

ruuskas: This should be the case only in `spectral_connectivity_epochs` (where multitaper frequencies must be defined by `rfftfreq` for the method to work). In `spectral_connectivity_time` there isn't really much reason to even use `rfftfreq` to obtain frequencies as it uses a different (time-resolved) multitaper implementation. If we require `freqs` anyway with `morlet`, it might make sense to go back to requiring them for `multitaper` as well given that, as far as I understand, in the time-resolved `tfr` implementation the difference between `morlet` and `multitaper` is only in the type of wavelet used (+ multiple tapers with multitaper).

ruuskas: Maybe a solution could be to add the connectivity paradigm (epochs vs time) to the parameters of `_compute_freqs`. Although the reasoning in my previous comment would still favor requiring `freqs` in Multitaper mode as well.

adam2392: Sure, I think requiring `freqs` could be fine. I'm not 100% sure I'm following wrt all the different possible combinations of parameters here, so I can take a look after you push up the proposed change?

mne_connectivity/spectral/tests/test_spectral.py
adam2392: Sounds good to me. I think we're allowed to make breaking changes as long as the results work rn. The package is not stable by any means.

adam2392:
Feel free to ping me whenever you need me to take a look again!

adam2392:
Happy new years @ruuskas ! Just wanted to check in on this. It would be great to get this in and make a new release w/ all your cool improvements!

ruuskas:
Happy new year @adam2392 ! I have been focusing on other things so this has been lagging. I'll try to get the necessary changes done this week (maybe today if there are no blockers).

ruuskas:
Hi @adam2392 ! This should now be ready for review again. The `freqs` parameter is now required and I added a test for padding. It was a good idea since there was a bug hiding indeed.

Sphinx is not happy, there seems to be an issue with indentation in the docstring, which I couldn't figure out without changing config.

drammock:
> Sphinx is not happy as there seems to be an issue with indentation in the docstring, which I couldn't figure out without changing config.

The sphinx error:

```
sphinx.errors.ThemeError: An error happened in rendering the page api.
Reason: UndefinedError("'logo' is undefined")
```
is due to a change in sphinx 6.0 and is currently being addressed upstream in the website theme. Temporary workaround is to pin sphinx to <6.0

ruuskas:
The Sphinx error I get locally is different. I can copy it verbatim tomorrow, but essentially "unexpected indentation" in the docstring. Indentation looks fine where it points to though.

mne_connectivity/spectral/time.py
```diff
@@ -431,23 +433,31 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
             window_length = np.arange(0., n_c / float(f), 1.0 / sfreq).shape[0]
             half_nbw = mt_bandwidth / 2.
             n_tapers = int(np.floor(mt_bandwidth - 1))
-            _, eigvals = dpss_windows(window_length, half_nbw, n_tapers,
-                                      sym=False)
+            _, eigvals = dpss_windows(window_length, half_nbw, n_tapers)
```
adam2392, line 436: Why do we remove `sym=False`?

mne_connectivity/spectral/time.py
```diff
@@ -458,137 +468,111 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
...
+                      s_yy.mean(axis=-1, keepdims=True))
+    coh = con_num / con_den
+    return coh
 
-    # define the function to compute in parallel
-    parallel, p_fun, n_jobs = parallel_func(
-        pairwise_cs, n_jobs=n_jobs, verbose=verbose, total=total)
 
-    return parallel(p_fun(s, t) for s, t in zip(source_idx, target_idx))
+def _cs(s_xy):
```
adam2392, line 574: Apparently, this line is not covered in the tests. I don't see it in the docstring either. What does `cs` stand for? If it's another connectivity function, can you add the relevant section in the docstring.

mne_connectivity/spectral/time.py
```diff
@@ -9,46 +9,50 @@
...
-    All methods are based on estimates of the cross- and power spectral
-    densities (CSD/PSD) Sxy and Sxx, Syy.
+    All methods are based on time-resolved estimates of the cross- and
+    power spectral densities (CSD/PSD) Sxy and Sxx, Syy.
 
     Parameters
     ----------
     data : array_like, shape (n_epochs, n_signals, n_times) | Epochs
         The data from which to compute connectivity.
+    freqs : array_like
```
adam2392, line 41: It can be `int`, or `float` right? If so, what is the expected behavior? It just computes the connectivity at that frequency?

mne_connectivity/spectral/time.py
```diff
@@ -184,6 +185,13 @@ def spectral_connectivity_time(data, method='coh', average=False,
 
             PLV = |E[Sxy/|Sxy|]|
 
+        'ciplv' : Corrected imaginary PLV (icPLV)
+        :footcite:`BrunaEtAl2018` given by::
```
adam2392, line 189:
```suggestion
        'ciplv' : Corrected imaginary PLV (icPLV) :footcite:`BrunaEtAl2018` given by::
```

mne_connectivity/spectral/time.py
```diff
@@ -82,6 +85,9 @@ def spectral_connectivity_time(data, method='coh', average=False,
         is equivalent to no smoothing.
     sm_kernel : {'square', 'hanning'}
         Smoothing kernel type. Choose either 'square' or 'hanning'.
+    padding : float
```
adam2392, line 88: Sorry maybe I missed this, but it's not clear to me what padding is meant to do. Do you mind adding a bit more info here in the docstring, or the `Notes` section?

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -613,6 +614,62 @@ def test_spectral_connectivity_time_resolved(method, mode):
                    for idx, jdx in triu_inds)
 
 
+@pytest.mark.parametrize('method', ['coh', 'plv', 'pli', 'wpli'])
+@pytest.mark.parametrize(
+    'mode', ['cwt_morlet', 'multitaper'])
+@pytest.mark.parametrize('padding', [0, 1, 5])
+def test_spectral_connectivity_time_padding(method, mode, padding):
+    """Test time-resolved spectral connectivity."""
```
adam2392, line 622:
```suggestion
    """Test time-resolved spectral connectivity with padding."""
```

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -613,6 +614,62 @@ def test_spectral_connectivity_time_resolved(method, mode):
...
+        fstart=fstart, fend=fend, trans_bandwidth=trans_bandwidth)
+    ch_names = np.arange(n_signals).astype(str).tolist()
+    info = create_info(ch_names=ch_names, sfreq=sfreq, ch_types='eeg')
+    data = EpochsArray(data, info)
+
+    # define some frequencies for tfr
+    freqs = np.arange(3, 20.5, 1)
+
+    # run connectivity estimation
+    if padding == 5:
```
adam2392, line 644: Why does padding == 5 not work?

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -613,6 +614,62 @@ def test_spectral_connectivity_time_resolved(method, mode):
...
+        with pytest.raises(ValueError):
```
adam2392, line 645: What is the specific ValueError that should be raised? e.g. `with pytest.raises(ValueError, match='...')`

adam2392 (changes requested):
LGTM! Some minor changes and clean up and this should be good to go soon.

mne_connectivity/spectral/time.py
ruuskas: Good point. I checked and it seems that `sym` was introduced in MNE-Python 1.3. I agree that it should be set to `False` to be more correct. Probably erroneously removed it due to some environment having MNE-Python < 1.3 and it somehow made it here.

adam2392:
@ruuskas I fixed the indentation error and other sphinx issue.

ruuskas:
Hi @adam2392! I addressed all the things you pointed out and pushed changes. WDYT?

mne_connectivity/spectral/time.py
```diff
@@ -82,6 +84,9 @@ def spectral_connectivity_time(data, method='coh', average=False,
...
+        Amount of time to consider as padding at the beginning and end of each
+        epoch in seconds.
```
adam2392, line 89:
```suggestion
        epoch in seconds. See Notes for more information.
```

mne_connectivity/spectral/time.py
```diff
@@ -458,137 +474,107 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
...
+        my_pairwise_con(w, psd, s, t, method, kernel,
+                        foi_idx, faverage, weights)
+        for s, t in zip(source_idx, target_idx))
 
 
-def _pli(w, kernel, foi_idx, source_idx, target_idx, n_jobs, verbose, total,
-         faverage, weights):
-    """Pairwise phase-lag index.
+def _pairwise_con(w, psd, x, y, method, kernel, foi_idx,
+                  faverage, weights):
```
adam2392, line 514: Even tho this is a private function, do you mind adding a basic docstring mainly for internal documentation purposes?

adam2392 (approved):
One minor nitpick, and then this LGTM for merging!

adam2392:
Thanks @ruuskas! Great addition and contribution.

ruuskas:
Thanks @adam2392!

PR #138, opened 2023-04-23
[ENH, WIP] Add multivariate connectivity methods

tsbinns:
@adam2392 Sorry for the delay, but I finally have something I think we can move forward with.

### **What I have implemented?**
- **Added multivariate measures of the imaginary part of coherency** (maximised imaginary coherence, MIC, and multivariate interaction measure, MIM; this includes spatial patterns of connectivity for MIC).
- **Added multivariate spectral Granger causality (GC).** Also made a slight change to the implementation which keeps the final results the same, but significantly cleans up the code.
- **These methods have been implemented in the existing `spectral_connectivity_epochs` function**, with the original structure of e.g. the `indices` parameter (i.e. no support for ragged indices added yet, as we discussed).
- **Updated the documentation of `spectral_connectivity_epochs` to reflect these new measures.**
- **Added new attributes to the connectivity classes.** These are used to store information associated with the multivariate measures (`rank`, `n_lags`, and `patterns`).
- **Support for parallelisation of these methods using the MNE wrapper for joblib is present.**
- **Added a (basic) set of unit tests for the new measures.**
- **Extensive examples for the newly-added measures.** These include background information/explanations of the methods, demonstrations of how to use the various settings in the implementations, how to interpret the results, what the limitations of these methods are, etc...

### **What are the limitations?**
- **Only bivariate **or** multivariate methods can be requested at any one time.** This relates to the existing `spectral_connectivity_epochs` function making assumptions about the number of signals in the seeds/targets matching the number of connections, which is not the case for the multivariate methods at this time. It is quite easy (and clean) to adapt the functions to handle this differently when only multivariate methods are called, however things become much trickier when trying to work with bivariate and multivariate methods simultaneously. If you would like the ability to compute bivariate and multivariate methods simultaneously to be included, perhaps we can adapt some of the functions to make this easier, otherwise we could stick with the current limitation I have imposed.
- **Only one multivariate connection can be computed at any one time.** This is not possible without changing the stucture of `indices`.
- **When computing multivariate connectivity, the number of seeds and targets for this single connection must be identical.** Again, this is not possible without allowing `indices` be ragged.
- **Support for computing GC on frequency bands is not yet possible.** This relates to the cross-frequency nature of GC measures. I had a solution which I had implemented in the code for my previous PR, however I think it would be good to discuss this particular issue before proceeding.

### **What have I not added?**
- **No unit tests for checking whether the methods are "correct".** I remember we discussed this in terms of adding some simulated data with a known connectivity distribution and testing the methods using this data. This is still on my to-do list; I just need to figure out how I could best do this with the smallest amount of data possible (so as not to bloat the package).
- **No version of the methods in `spectral_connectivity_time`.** Again, this is still on my to-do list. I have structured the MIC/MIM and GC computation code in a way that we could very easily re-use the code I have added here and just switch out the time dimension (used in case of time-frequency modes) to store epoch information, so I do not consider this a major hurdle.

**There is, however, a bigger issue:** the assertion of the connectivity object in `test_spectral_connectivity_parallel` matching that which has been saved and then reloaded now fails, because there is ~1 kB size difference in the objects. As far as I can tell, all of the critical information in the objects are identical (e.g. the results are the same, all entries in `attrs` are the same). Interestingly, this only occurs for the multitaper and fourier modes. I checked in the original test, and the size of the saved and reloaded connectivity object is also not a perfect match, however it seems to be rounded to the same "~XX kB". My hunch would be that the attrs I added to the connectivity classes (`rank`, `n_lags`, and `patterns`), even when unfilled, are pushing the estimated size in the repr to be rounded up when combined with the pre-existing small difference in size that occurs when saving and reloading a connectivity object. Even though I think everything is in order, I am clearly not happy with this test suddenly failing, so if you have any ideas as to why this is happening, I would really appreciate it!

**The test in question:**
https://github.com/tsbinns/mne-connectivity/blob/100e235c8520664510401d73394df1c953a6721f/mne_connectivity/spectral/tests/test_spectral.py#L134-L145

**An example of the reprs from the test with the multitaper mode:**
image

### **What are the next steps?**
Adding further unit tests, as well as implementing these measures in `spectral_connectivity_time` are my two biggest targets. Once I have your input, I am happy to also start addressing the limitations I listed (e.g. by adding support for ragged indices). Making sure all tests are passing would also be nice!

One again, I am very sorry for the delay (a lot of other work got in the way), but I am still very excited to move forward with this! If there is anything you think would be best discussed over a call, I am very happy to do that again. I will also answer anything here as soon as I can.

Cheers,
Thomas


mne_connectivity/base.py
```diff
@@ -645,6 +650,16 @@ def n_epochs_used(self):
...
+    @property
+    def rank(self):
+        """Seed and target ranks used in the connectivity."""
+        return self.attrs['rank']
+
+    @property
+    def n_lags(self):
+        """Number of lags used in the model for computing connectivity."""
+        return self.attrs['n_lags']
+
```
adam2392, line 662: For these additional properties + `patterns`, I would be inclined to put these altogether into a new class for `MultivariateMixin`.

Slash, do we necessarily need these? That is we don't need to expose all `attrs`. Rn it exposes the core ones, but presumably a user can store arbitrary data fields in there.

mne_connectivity/spectral/epochs.py
```diff
@@ -222,6 +246,8 @@ def compute_con(self, con_idx, n_epochs):
 class _EpochMeanConEstBase(_AbstractConEstBase):
     """Base class for methods that estimate connectivity as mean epoch-wise."""
 
+    patterns = None
```
adam2392, line 249: What is `patterns`?

mne_connectivity/spectral/epochs.py
```diff
@@ -1123,39 +1871,60 @@ def spectral_connectivity_epochs(data, names=None, method='coh', indices=None,
...
         this_con = conn_method.con_scores
+        this_patterns = conn_method.patterns
 
-        if this_con.shape[0] != n_cons:
-            raise ValueError('First dimension of connectivity scores must be '
-                             'the same as the number of connections')
+        assert this_con.shape[0] == n_cons, (
+            'first dimension of connectivity scores does not match the '
+            'number of connections; please contact the mne-connectivity '
+            'developers')
```
adam2392, line 1899: We generally don't want `assert` messages as these do not get run if the Python package binary is distributed.

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -408,6 +408,137 @@ def test_spectral_connectivity(method, mode):
     assert (out_lens[0] == 10)
 
 
+@pytest.mark.parametrize('method', ['mic', 'mim', 'gc', 'gc_tr',
+                                    ['mic', 'mim', 'gc', 'gc_tr']])
+@pytest.mark.parametrize('mode', ['multitaper', 'fourier', 'cwt_morlet'])
+def test_multivar_spectral_connectivity(method, mode):
+    """Test multivariate frequency-domain connectivity methods."""
```
adam2392, line 415: Is this mainly a test to catch errors that should be raised? If so, can we name the test and add a note in the docstring accordingly?

adam2392 (commented):
Thanks for opening the PR!

I will take a look later this week. I'm not exactly sure what `patterns` and some of the added attributes are. I left some other basic comments upon skimming this.

Do you mind adding some more details summarizing them here?

mne_connectivity/base.py
tsbinns: I only included them here so that they would be included in the docs for the connectivity classes.
Something like `patterns` I would be more inclined to expose, since these are relevant results of the class, however I see no issue not exposing `rank` and `n_lags`.
Perhaps I could write a `MultivariateMixin` for `patterns`, or do you think it would be a bit overkill to only expose one attribute?

mne_connectivity/spectral/tests/test_spectral.py
tsbinns: At the moment, yes, this is really just catching errors (since much of the behaviour is already being tested in `test_spectral_connectivity`). I will update the name and description.

mne_connectivity/base.py
tsbinns: `patterns` are the spatial topographies I mentioned in the previous PR (generated from the spatial filters used to compute MIC). They can also be called spatial patterns, and this "patterns" terminology is more consistent with other parts of MNE, e.g. [mne.decoding.SSD](https://mne.tools/dev/generated/mne.decoding.SSD.html).

mne_connectivity/spectral/epochs.py
tsbinns: **FIXED**

adam2392:
Sorry for the delay. I fixed any irrelevant CI issues, but the rest it seems stems from changes in this PR. @tsbinns

mne_connectivity/base.py
adam2392: Ah I see.

If they are only used for `MIC`, then I would just document the `MIC` function stating that `patterns` is available in the `attrs` object. Perhaps even explicitly showing this in the example should suffice too. Otherwise it would be weird to have a property always be set to `None` unless MIC is used.

tsbinns:
@adam2392 I have been continuing to add the changes/fixes discussed above.

The main thing now would be how to handle the ragged arrays that could occur when working with multivariate methods (e.g. indices, results, patterns). We discussed some ideas for this, but never came to a decision.
Would you prefer it if I add some MultivariateMixIn class, or would you want something more general like an N-dimensional results class of its own that could be used for other purposes as well? I think that would be the last big step before this can be finalised.

Apologies for the delays in sorting this; unfortunately I have not been able to dedicate as much time as I would have liked to the PR, but it still something I am very much interested in!

adam2392:
> The main thing now would be how to handle the ragged arrays that could occur when working with multivariate methods (e.g. indices, results, patterns). We discussed some ideas for this, but never came to a decision. Would you prefer it if I add some MultivariateMixIn class, or would you want something more general like an N-dimensional results class of its own that could be used for other purposes as well? I think that would be the last big step before this can be finalised.

Let's revisit this when this PR is merged if that's okay with you? IIUC, we don't need the ragged part for all the methods, and thus, per #125 , we can implement this first?

I think the ragged part would take some discussion and ideally some thought to make it as lightweight as possible and compatible with where scientific python is heading: https://discuss.scientific-python.org/t/ragged-array-summit/465/4. We might consider adding an optional dependency, implementing the soln. ourselves, or a combination. 

> Apologies for the delays in sorting this; unfortunately I have not been able to dedicate as much time as I would have liked to the PR, but it still something I am very much interested in!

No problem! I've been quite swamped. Ping me whenever you have the changes/fixes up.

tsbinns:
**DONE:**

- removed the explicit attributes for multivariate info, now just stored in `attrs`
- general improvements to make the examples clearer
- general documentation updates
- fixed the bugs that were causing tests to fail

**IN PROGRESS:**

- adding the methods to `spectral_connectivity_time`

**OUTSTANDING ISSUES:**

- tests on the similarity of the connectivity objects after saving still failing due to 1 kB difference in the object sizes despite results being the same; same thing I mentioned in the first post:

> The assertion of the connectivity object in `test_spectral_connectivity_parallel` matching that which has been saved and then reloaded now fails, because there is ~1 kB size difference in the objects. As far as I can tell, all of the critical information in the objects are identical (e.g. the results are the same, all entries in `attrs` are the same). Interestingly, this only occurs for the multitaper and fourier modes. I checked in the original test, and the size of the saved and reloaded connectivity object is also not a perfect match, however it seems to be rounded to the same "~XX kB". Even though I think everything is in order, I am clearly not happy with this test suddenly failing, so if you have any ideas as to why this is happening, I would really appreciate it!
> 
> **The test in question:** https://github.com/tsbinns/mne-connectivity/blob/100e235c8520664510401d73394df1c953a6721f/mne_connectivity/spectral/tests/test_spectral.py#L134-L145
> 
> **An example of the reprs from the test with the multitaper mode:** image

I'm not sure why this is happening, so I would appreciate any input if you have the time @adam2392.

adam2392:
> * adding the methods to `spectral_connectivity_time`

Hi @tsbinns how is this part going? Would you like me to review anything? Thanks!

tsbinns:
> > * adding the methods to `spectral_connectivity_time`
> 
> Hi @tsbinns how is this part going?

Hi @adam2392, this is still a WIP, however from Monday I have 2 full weeks to dedicate to this, so it will be finished then. I will tag you once I have it and all tests are passing.


> Would you like me to review anything?

I would be interested to hear your thoughts on the examples I added (`mic_mim.py` and `granger_causality.py`) and documentation changes I made for `spectral_connectivity_epochs`. Do you think there are any improvements I could make? Cheers!

tsbinns:
Hi @adam2392, I have now:
- added support for the methods in _spectral_connectivity_time_
- added tests for the methods in both _spectral_connectivity_epochs_ and _time_ (including a regression test for some example data for comparing results against the established MATLAB implementation of the methods)
- fixed some bugs which I had previously missed

In terms of the goals of this PR I believe that is everything finished, assuming it is up to standard. I would be very grateful for any feedback!

examples/granger_causality.py
```diff
@@ -0,0 +1,405 @@
...
+# causality (GC). A signal, :math:`\boldsymbol{x}`, is said to Granger-cause
+# another signal, :math:`\boldsymbol{y}`, if information from the past of
+# :math:`\boldsymbol{x}` improves the prediction of the present of
+# :math:`\boldsymbol{y}` over the case where only information from the past of
+# :math:`\boldsymbol{y}` is used. Note: GC does not make any assertions about
+# the true causality between signals.
+#
+# The degree to which :math:`\boldsymbol{x}` and :math:`\boldsymbol{y}` can be
+# used to predict one another can be quantified using vector autoregressive
+# (VAR) models. Considering the simpler case of time domain connectivity, the
```
adam2392, line 49:
```suggestion
# The degree to which :math:`\boldsymbol{x}` and :math:`\boldsymbol{y}` can be
# used to predict one another in a linear model can be quantified using vector autoregressive
# (VAR) models. Considering the simpler case of time domain connectivity, the
```

examples/granger_causality.py
```diff
@@ -0,0 +1,405 @@
...
+#
+# and :math:`\boldsymbol{z}_t = \sum_{k=1}^K \boldsymbol{A}_k
+# \boldsymbol{z}_{t-k} + \boldsymbol{\epsilon}_t` ,
+# :math:`\boldsymbol{\Sigma} := \langle \boldsymbol{\epsilon}_t
+# \boldsymbol{\epsilon}_t^T \rangle = \begin{bmatrix} \Sigma_{xx} & \Sigma_{xy}
+# \\ \Sigma_{yx} & \Sigma_{yy} \end{bmatrix}` ,
+#
+# representing the reduced and full VAR models, respectively, where: :math:`K`
+# is the order of the VAR model, determining the number of lags, :math:`k`,
+# used; :math:`\boldsymbol{Z} := \begin{bmatrix} \boldsymbol{x} \\
```
adam2392, line 63:
```suggestion
# used; :math:`\boldsymbol{A}` is the array (e.g. a 2D matrix) of coefficients we want to estimate;:math:`\boldsymbol{Z} := \begin{bmatrix} \boldsymbol{x} \\
```

examples/granger_causality.py
```diff
@@ -0,0 +1,405 @@
...
+signals_b_names = [epochs.info['ch_names'][idx] for idx in signals_b]
+
+# compute Granger causality
+gc_ab = spectral_connectivity_epochs(
+    epochs, method=['gc'], indices=indices_ab, fmin=5, fmax=30,
+    rank=(np.array([5]), np.array([5])), gc_n_lags=20)  # A => B
+gc_ba = spectral_connectivity_epochs(
+    epochs, method=['gc'], indices=indices_ba, fmin=5, fmax=30,
+    rank=(np.array([5]), np.array([5])), gc_n_lags=20)  # B => A
+freqs = gc_ab.freqs
```
adam2392, line 170: Can you write a note here explaining the intuition of `rank` and `gc_n_lags`? I know they are in the docstring, but perhaps for a user that is used to the original API of `spectral_connectivity_epochs`, a short note here might help them understand how to set the values as well.

adam2392: And perhaps the same in the `mic_mim.py` example

mne_connectivity/spectral/epochs.py
```diff
@@ -89,14 +90,36 @@ def _prepare_connectivity(epoch_block, times_in, tmin, tmax,
...
-        # only compute r for lower-triangular region
-        indices_use = np.tril_indices(n_signals, -1)
+        if any(this_method in _multivariate_methods for this_method in method):
+            if any(this_method in _gc_methods for this_method in method):
+                raise ValueError(
+                    'indices must be specified when computing Granger '
+                    'causality, as all-to-all connectivity is not supported')
+            else:
+                logger.info('using all indices for multivariate connectivity')
+                indices_use = (np.arange(n_signals, dtype=int),
```
adam2392, line 100: Can you add a unit-test for this case?

mne_connectivity/spectral/epochs.py
```diff
@@ -89,14 +90,36 @@ def _prepare_connectivity(epoch_block, times_in, tmin, tmax,
...
+                raise ValueError(
+                    'seed and target indices must not intersect when computing'
+                    'Granger causality')
         indices_use = check_indices(indices)
 
     # number of connectivities to compute
-    n_cons = len(indices_use[0])
+    if any(this_method in _multivariate_methods for this_method in method):
+        if len(np.unique(indices_use[0])) != len(np.unique(indices_use[1])):
+            raise ValueError(
```
adam2392, line 117: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -443,19 +582,33 @@ def _spectral_connectivity(data, method, kernel, foi_idx,
...
+        Dictionary containing the connectivity estimates corresponding to the
+        metrics in ``method``. Each element is an array of shape (n_pairs,
+        n_freqs) or (n_pairs, n_fbands) if ``faverage`` is `True`.
+
+    patterns : dict
+        Dictionary containing the connectivity patterns corresponding to the
+        metrics in ``method``, if multivariate methods are called, else an
+        empty dictionary. Each element is an array of shape (n_pairs,
+        n_channels, n_freqs) or (n_pairs, n_channels, n_fbands) if ``faverage``
+        is `True`.
```
adam2392, line 611: Sorry I'm not familiar. Can you elaborate what each element represents here?

From https://mne.tools/dev/generated/mne.decoding.SSD.html, I get `The patterns for reconstructing the signal from the filtered data.`, but I presume that's not exactly what it is here? 

Just want to make sure it's crystal clear :p.

mne_connectivity/spectral/time.py
```diff
@@ -639,6 +816,96 @@ def _pairwise_con(w, psd, x, y, method, kernel, foi_idx,
...
+        patterns.append(con_est.patterns)
+        if patterns[-1] is not None:
+            patterns[-1] = patterns[-1][..., np.newaxis]
+
+    for i, _ in enumerate(scores):
+        # mean inside frequency sliding window (if needed)
+        if isinstance(foi_idx, np.ndarray) and faverage:
+            scores[i] = _foi_average(scores[i], foi_idx)
+            if patterns[i] is not None:
+                patterns[i] = _foi_average(patterns[i], foi_idx)
```
adam2392, line 900: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -385,18 +512,28 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
...
+        store_params = {
+            'data': conn[m], 'patterns': patterns[m], 'freqs': out_freqs,
+            'n_nodes': n_signals, 'names': names, 'indices': indices,
+            'method': method, 'spec_method': mode, 'events': events,
+            'event_id': event_id, 'metadata': metadata, 'rank': rank,
+            'n_lags': gc_n_lags if m in _gc_methods else None}
+        if average:
+            store_params['data'] = np.mean(store_params['data'], axis=0)
+            if patterns[m] is not None:
+                store_params['patterns'] = np.mean(store_params['patterns'],
```
adam2392, line 532: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -302,12 +391,42 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
...
+            indices_use = np.tril_indices(n_signals, k=-1)
     else:
+        if multivariate_con:
+            if len(np.unique(indices[0])) != len(np.unique(indices[1])):
+                raise ValueError(
+                    'seed and target indices cannot contain repeated '
+                    'channels for multivariate connectivity')
+            if any(this_method in _gc_methods for this_method in method):
+                if set(indices[0]).intersection(indices[1]):
+                    raise ValueError(
```
adam2392, line 413: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -302,12 +391,42 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
...
+            logger.info('using all indices for multivariate connectivity')
+            indices_use = (np.arange(n_signals, dtype=int),
+                           np.arange(n_signals, dtype=int))
+        else:
+            logger.info('only using indices for lower-triangular matrix')
+            indices_use = np.tril_indices(n_signals, k=-1)
     else:
+        if multivariate_con:
+            if len(np.unique(indices[0])) != len(np.unique(indices[1])):
+                raise ValueError(
```
adam2392, line 408: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -302,12 +391,42 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
 
     # get indices of pairs of (group) regions
     if indices is None:
-        indices_use = np.tril_indices(n_signals, k=-1)
+        if multivariate_con:
+            if any(this_method in _gc_methods for this_method in method):
+                raise ValueError(
```
adam2392, line 396: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -284,6 +357,22 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
...
+    if len(fmin) != 1 and any(
+        this_method in _gc_methods for this_method in method
+    ):
+        raise ValueError('computing Granger causality on multiple frequency '
+                         'bands is not yet supported')
+
+    if any(this_method in _multivariate_methods for this_method in method):
+        if not all(this_method in _multivariate_methods for
+                   this_method in method):
+            raise ValueError(
```
adam2392, line 369: Can you add a unit-test for this case?

mne_connectivity/spectral/time.py
```diff
@@ -284,6 +357,22 @@ def spectral_connectivity_time(data, freqs, method='coh', average=False,
     if np.any(fmin > fmax):
         raise ValueError('fmax must be larger than fmin')
 
+    if len(fmin) != 1 and any(
+        this_method in _gc_methods for this_method in method
+    ):
+        raise ValueError('computing Granger causality on multiple frequency '
```
adam2392, line 363: Can you add a unit-test for this case?

adam2392 (commented):
Hi @tsbinns sorry for the delay! 

I left some notes. This is looking good imo! The examples are a nice touch because for someone that doesn't understand what these methods were previously, I feel like I'm learning something.

The main nits I have are unit-test coverage. You can just add separate unit-tests to achieve these. Other things I reviewed are just nits on explaining a concept and documentation.

I think once these are addressed, maybe a few more small changes. Overall though the high-level is fine w/ me and once it's in the hands of users for v0.6, we'll see how it goes :)

examples/granger_causality.py
```diff
@@ -0,0 +1,405 @@
...
+# :math:`\boldsymbol{\epsilon}` are the residuals of the VAR models. In this
+# way, the information of the signals at time :math:`t` can be represented as a
+# weighted form of the information from the previous timepoints, plus some
+# residual information not encoded in the signals' past. In practice, VAR model
+# parameters are computed from an autocovariance sequence generated from the
+# time-series data using the Yule-Walker equations :footcite:`Whittle1963`.
+#
+# By comparing the residuals, or errors, of the reduced and full VAR models, we
+# can therefore estimate how much :math:`\boldsymbol{x}` Granger-causes
+# :math:`\boldsymbol{y}`:
```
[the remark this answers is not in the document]
tsbinns: Sort of. It would be more the variance of the residuals of the (reduced) model based on the information in y alone vs. the variance of the residuals of the (full) model based on the information in both y and x. I have re-written this part to try and make it clearer.

examples/granger_causality.py
tsbinns: For `gc_n_lags`, I explain this in the section below (applies only to GC, not to MIC/MIM): https://github.com/tsbinns/mne-connectivity/blob/dae09a5ff3a1bd5caf848d8ab711d1f3887280ad/examples/granger_causality.py#L309-L319

As for `rank`, this applies to GC and MIC/MIM, so I explain this in the section below for the GC example: https://github.com/tsbinns/mne-connectivity/blob/dae09a5ff3a1bd5caf848d8ab711d1f3887280ad/examples/granger_causality.py#L342-L382
... and for the MIC/MIM example: https://github.com/tsbinns/mne-connectivity/blob/dae09a5ff3a1bd5caf848d8ab711d1f3887280ad/examples/mic_mim.py#L293-L326

mne_connectivity/base.py
```diff
@@ -685,6 +686,10 @@ def get_data(self, output='compact'):
         if output == 'raveled':
             data = self._data
         else:
+            if self.method in ['mic', 'mim', 'gc', 'gc_tr']:
+                raise ValueError('cannot return multivariate connectivity '
```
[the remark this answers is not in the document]
tsbinns: So at the moment it's because the results for a single connection would correspond to multiple entries in the matrix (for the various channels involved), and if we extend the functionality to compute multiple connections in a single function call, we could end up with a case where the results for multiple connections correspond to the same matrix entries (i.e. results for 0 -> 1 would overlap with results for 0 -> [1, 2]). I have added a comment stating this.

mne_connectivity/spectral/tests/data/example_multivariate_data.pkl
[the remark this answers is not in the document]
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/data/README.md

mne_connectivity/spectral/time.py
tsbinns: So it's very similar, but instead of reconstructing the signal we're reconstructing the largest connectivity component of the signal. I have updated to try and make it clearer:
https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/time.py#L613-L619

mne_connectivity/spectral/time.py
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/test_spectral.py#L1050-L1088

mne_connectivity/spectral/time.py
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/test_spectral.py#L1050-L1088

mne_connectivity/spectral/time.py
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/test_spectral.py#L1103-L1108

mne_connectivity/spectral/time.py
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/test_spectral.py#L1132-L1135

mne_connectivity/spectral/epochs.py
tsbinns: Done! https://github.com/tsbinns/mne-connectivity/blob/894c42e59cef9f97aa07a7a4f71cef756512c50c/mne_connectivity/spectral/tests/test_spectral.py#L593-L599

tsbinns:
Thanks very much for the feedback @adam2392! I believe I have addressed everything, but please unresolve any comments you think I should work on further.

Also please let me know when you think it's appropriate for me to update the `whats_new.rst`.

Cheers!

adam2392:
@tsbinns feel free to add a changelog entry rn that summarizes this major feature effort! Follow the pattern there to link your name/github page.

I will take a look later in-depth on the code.

examples/mic_mim.py
```diff
@@ -0,0 +1,410 @@
...
+# %%
+
+# left hemisphere sensors
+seeds = [idx for idx, ch_info in enumerate(epochs.info['chs']) if
+         ch_info['loc'][0] < 0]
+# right hemisphere sensors
+targets = [idx for idx, ch_info in enumerate(epochs.info['chs']) if
+           ch_info['loc'][0] > 0]
+
+# UNTIL RAGGED INDICES SUPPORTED
```
adam2392, line 84:
```suggestion
# XXX: Currently ragged indices are not supported, so we only consider a single list of indices
```

adam2392: Is this right?

examples/mic_mim.py
```diff
@@ -0,0 +1,410 @@
...
+
+###############################################################################
+# Furthermore, spatial patterns of connectivity can be constructed from the
+# spatial filters to give a picture of the location of the sources involved in
+# the connectivity. This information is stored under ``attrs['patterns']`` of
+# the connectivity class, with one value per frequency for each channel in the
+# seeds and targets. As with MIC, the absolute value of the patterns reflect
+# the strength, however the sign differences can be used to visualise the
+# orientation of the underlying dipole sources. The spatial patterns are not
+# bound between :math:`[-1, 1]`.
```
adam2392, line 169:
```suggestion
# orientation of the underlying dipole sources. Note the spatial patterns are **not**
# bound between :math:`[-1, 1]`.
```

examples/mic_mim.py
```diff
@@ -0,0 +1,410 @@
...
+# MIM. MIM can be thought of as an average of all connectivity components
+# between the seeds and targets, and can be useful for an exploration of all
+# available components. It is unnecessary to use the spatial filters of each
+# component explicitly, and instead the desired result can be achieved from
+# :math:`E` alone (Eq. 14 of :footcite:`EwaldEtAl2012`):
+#
+# :math:`MIM=tr(\boldsymbol{EE}^T)`,
+#
+# where again the frequency dependence is omitted. Unlike MIC, MIM is
+# positive-valued, and is not bound below 1. Without normalisation, MIM can be
```
adam2392, line 237: What does "not bound below 1" mean?

examples/mic_mim.py
```diff
@@ -0,0 +1,410 @@
...
+# thought of as reflecting the total interaction between the seeds and targets.
+# MIM can be normalised to lie in the range :math:`[0, 1]` by dividing the
+# scores by the number of channels in the seeds and targets, representing the
+# total interaction *per channel*, which can accordingly be influenced by e.g.
+# the presence of channels with little to no interaction. **MIM is not
+# normalised by default in this implementation**.
```
adam2392, line 243: Perhaps we can add a few LOC showing what a normalization would look like of `mim`?

examples/granger_causality.py
```diff
@@ -0,0 +1,413 @@
...
+# One important parameter when computing GC is the number of lags used when
+# computing the VAR model. A lower number of lags reduces the computational
+# cost, but in the context of spectral GC, leads to a smoothing of Granger
+# scores across frequencies. The number of lags can be specified using the
+# ``gc_n_lags`` parameter. The default value is 40, however there is no correct
+# number of lags to use when computing GC. Instead, you have to use your own
+# best judgement of whether or not your Granger scores look overly smooth.
+#
+# Below is a comparison of Granger scores computed with a different number of
+# lags. In the above examples we had used 20 lags, which we will compare to
```
adam2392, line 322:
```suggestion
# lags. In the above examples we used 20 lags, which we will compare to
```

examples/granger_causality.py
```diff
@@ -0,0 +1,413 @@
...
+# %%
+
+# parietal sensors
+signals_a = [idx for idx, ch_info in enumerate(epochs.info['chs']) if
+             ch_info['ch_name'][2] == 'P']
+# occipital sensors
+signals_b = [idx for idx, ch_info in enumerate(epochs.info['chs']) if
+             ch_info['ch_name'][2] == 'O']
+
+# UNTIL RAGGED INDICES SUPPORTED
```
adam2392, line 160:
```suggestion
# XXX: Currently ragged indices are not supported, so we only consider a single list of indices
```

adam2392 (approved):
I think this LGTM once some final nits on grammar is fixed.

examples/mic_mim.py
tsbinns: MIM can be > 1. Rewritten to "Unlike MIC, MIM is positive-valued and can be > 1."

examples/mic_mim.py
tsbinns: That's right, I just reworded slightly for clarity: https://github.com/mne-tools/mne-connectivity/blob/49dd83c04c50c5dc7133e8b8ee9ccb0ff716d1a0/examples/mic_mim.py#L84-L85

examples/mic_mim.py
tsbinns: How's this? https://github.com/mne-tools/mne-connectivity/blob/49dd83c04c50c5dc7133e8b8ee9ccb0ff716d1a0/examples/mic_mim.py#L271-L273

I also added some info. to make the interpretation of the normalised metrics clearer, e.g.:
https://github.com/mne-tools/mne-connectivity/blob/49dd83c04c50c5dc7133e8b8ee9ccb0ff716d1a0/examples/mic_mim.py#L259-L261

examples/granger_causality.py
tsbinns: Same rewording as above: https://github.com/mne-tools/mne-connectivity/blob/49dd83c04c50c5dc7133e8b8ee9ccb0ff716d1a0/examples/granger_causality.py#L160-L161

adam2392 (approved):

adam2392:
LGTM. Thanks @tsbinns and team!

drammock:
congrats on getting this one merged in!  great effort.

Issue #170, opened 2024-02-02
`spectral_connectivity_time` method list not rendering properly


tsbinns:
[`mne_connectivity.spectral_connectivity_time()`](https://mne.tools/mne-connectivity/dev/generated/mne_connectivity.spectral_connectivity_time.html) has a nice overview of the supported connectivity methods under the `methods` parameter:
https://github.com/mne-tools/mne-connectivity/blob/043d88873b968f948e9571237c5f4ef8e0a4cef4/mne_connectivity/spectral/time.py#L71-L85

<br>

Looking at the code I'm guessing this is supposed to be a list of bullet points, however it does not render this way:
![image](https://github.com/mne-tools/mne-connectivity/assets/56922019/552661f5-feca-487b-965f-7e9abc49aeb1)

For this to render properly, there should be a new line before and after the list.

**How would people feel about a small PR to add these new lines?**

<br>

Related point: I like this way of giving a quick overview of the supported methods. Do you think it would be worth adding the same in [`mne_connectivity.spectral_connectivity_epochs()`](https://mne.tools/mne-connectivity/dev/generated/mne_connectivity.spectral_connectivity_epochs.html)?

larsoner:
Yeah newlines would be good. +1 for adding to the `epochs` docstring as well, preferably by a deduplication / use of `docs.py`

tsbinns:
@larsoner Given that the methods are slightly different between `epochs` and `time`, would it still be possible to deduplicate this using `docs.py`?

larsoner:
Can you deduplicate *some but not all* of the methods? That might still be worth it. But if it's too much of a pain, duplicating the docstring is okay.

tsbinns:
Didn't realise `fill_doc` was so flexible. Will open a PR with the changes.

tsbinns:
Looking at the render on my end it doesn't seem to break the other formatting. Might have to incorporate this in some other projects!

Opened a PR in #171.

PR #171, opened 2024-02-05
[DOC] Add methods list to `epochs` function and fix formatting of list in `time` function

tsbinns:
Follow up from #170:

- Added newlines before and after methods list in `spectral_connectivity_time()` to ensure it renders properly.
- Added equivalent list to `spectral_connectivity_epochs()`.
- Added definition of individual methods to `docs.py`.

@larsoner Please let me know if the definition of strings in `docs.py` works for you.

Cheers!


larsoner (approved):
+1 for merge when green, so probably after #172 lands and we merge `main` into this PR

PR #163, opened 2023-12-12
[ENH] Add support for multivariate coherency method (canonical coherence)

orabe:
This PR adds support for another multivariate connectivity method (canonical coherence; CaCoh), taking advantage of the multivariate framework implemented in v0.6. It is authored by myself and @tsbinns.

Method paper: [Vidaurre *et al.* (2019)](https://www.sciencedirect.com/science/article/abs/pii/S1053811919305907)

As mentioned in #155, canonical coherence is a multivariate method based on coherency. It shares a number of similarities with the recently added methods MIC & MIM, but it is a more appropriate method to use for datasets where you don't expect volume conduction artefacts to be a big issue (e.g. connectivity between: cortical and subcortical recordings; cortex & EMG recordings, etc...).

Members of my own and other groups would be very interested in having this method implemented in MNE, as currently there is no implementation in any signal processing package (Python or MATLAB; only some MATLAB scripts provided by the authors on request).

## Changes

Because CaCoh and MIC/MIM are based on similar principles, several bits of code can be shared between them. Accordingly, we:

Refactored the `_MultivariateCohEstBase` class to accommodate all multivariate coherency methods (CaCoh, MIC, & MIM):

https://github.com/mne-tools/mne-connectivity/blob/f9ff2ac7f1baea1c71d989e909fa5efae3df6b04/mne_connectivity/spectral/epochs_multivariate.py#L179-L185

Introduced a new class `_MultivariateImCohEstBase` for code specific to MIC & MIM

https://github.com/mne-tools/mne-connectivity/blob/f9ff2ac7f1baea1c71d989e909fa5efae3df6b04/mne_connectivity/spectral/epochs_multivariate.py#L341-L346

Introduced a new class `_CaCohEst` for code specific to CaCoh

https://github.com/mne-tools/mne-connectivity/blob/f9ff2ac7f1baea1c71d989e909fa5efae3df6b04/mne_connectivity/spectral/epochs_multivariate.py#L470-L475

There are no API changes (beyond the ability to specify a new method).

## Unit Tests

Unit tests for CaCoh have been implemented similarly to the other multivariate methods, which in most cases was just the addition of the `"cacoh"` method parameter to existing tests.

Notably, a regression test based on results from MATLAB code of the implementation was added, as for the other multivariate methods.

## Outstanding Tasks

[...]

PR #173, opened 2024-03-01
[ENH] Add connectivity simulation function

tsbinns:
## Summary

Adds a new public function `make_signals_in_freq_bands()` inside `mne_connectivity/datasets/frequency.py` for a basic simulation of connectivity, following @adam2392's suggestion in #163.

<br>

## API/Documentation

The function has parameters for what I feel are the most important simulation options.

I added a new "Dataset functions" section to the API page since I wasn't sure how well it fit under the existing headers and link to the function here. Very open to change this if you have better suggestions.

<br>

## Unit tests

I added a new set of tests in `mne_connectivity/tests/test_datasets.py`.

`test_make_signals_error_catch()` checks that some obvious errors are caught. If you think there are some important checks I've missed please let me know.

`test_make_signals_in_freq_bands` checks that the simulations are actually working properly. I focused on Coh, ImCoh, and DPLI with: 1) no time delay in interaction between seeds and targets; 2) a positive time delay (i.e. seeds drive communication with targets); and 3) a negative time delay (i.e. targets drive communication with seeds). From this we expect:

- Case 1 (no time delay) - Coh should be high for the simulated frequency band, ImCoh should be low, DPLI should be ~0.5.
- Case 2 (positive delay) - Coh should be high, ImCoh should be high, DPLI should be > 0.5.
- Case 3 (negative delay) - Coh should be high, ImCoh should be high, DPLI should be < 0.5.

I test these with a higher and lower SNR, different number of seeds and targets, and the different CSD computation modes. I realise it's not an exhaustive check, but I think the tests nicely capture what we expect to see from undirected (Coh, ImCoh) and directed (DPLI) connectivity methods, and also methods that do not capture zero time-lag interactions (ImCoh).

<br>

## Use elsewhere in the package

This simulation function is similar (**but not identical**) to those used for many of the unit tests in `spectral.py`. Out of interest I  wanted to see if this function could be used to replace these simulations. In a number of cases it works, but in others it would require redefining what is considered acceptable connectivity values for particular method/CSD mode combinations.

This could be done, but it would require a considerable amount of effort tweaking these thresholds which I can't really invest the time for at the moment.

How would people feel about just having this new function which can be used going forward for e.g. the CaCoh examples in #163 without changing the existing unit test stimulations?

### DIFF
--- a/mne_connectivity/datasets/frequency.py
+++ b/mne_connectivity/datasets/frequency.py
@@ -43,7 +43,8 @@ def make_signals_in_freq_bands(
         Sampling frequency of the simulated data, in Hz.
     trans_bandwidth : int | float (default 1)
         Transition bandwidth of the filter to apply to isolate activity in
-        ``freq_band``, in Hz.
+        ``freq_band``, in Hz. These are passed to the ``l_bandwidth`` and ``h_bandwidth``
+        keyword arguments in :func:`mne.filter.create_filter`.
     snr : float (default 0.7)
         Signal-to-noise ratio of the simulated data in the range [0, 1].
     connection_delay : int (default 5)
@@ -54,7 +55,8 @@ def make_signals_in_freq_bands(
         Earliest time of each epoch.
     ch_names : list of str | None (default None)
         Names of the channels in the simulated data. If `None`, the channels are named
-        according to their index and the frequency band of interaction.
+        according to their index and the frequency band of interaction. If specified, must be a list of
+        ``n_seeds + n_targets`` channel names.
     ch_types : str | list of str (default "eeg")
         Types of the channels in the simulated data. If specified as a list, must be a
         list of ``n_seeds + n_targets`` channel names.
@@ -63,7 +65,7 @@ def make_signals_in_freq_bands(
 
     Returns
     -------
-    epochs : mne.EpochsArray
+    epochs : mne.EpochsArray of shape (n_epochs, n_seeds + n_targets, n_times)
         The simulated data stored in an `mne.EpochsArray` object. The channels are
         arranged according to seeds, then targets.
 
@@ -101,10 +103,10 @@ def make_signals_in_freq_bands(
         )
 
     # simulate data
-    rng = np.random.RandomState(rng_seed)
+    rng = np.random.default_rng(rng_seed)
 
     # simulate signal source at desired frequency band
-    signal = rng.randn(1, n_epochs * n_times + np.abs(connection_delay))
+    signal = rng.standard_normal(size=(1, n_epochs * n_times + np.abs(connection_delay)))
     signal = filter_data(
         data=signal,
         sfreq=sfreq,
@@ -116,7 +118,7 @@ def make_signals_in_freq_bands(
     )
 
     # simulate noise for each channel
-    noise = rng.randn(n_channels, n_epochs * n_times + np.abs(connection_delay))
+    noise = rng.standard_normal(size=(n_channels, n_epochs * n_times + np.abs(connection_delay)))
 
     # create data by projecting signal into each channel of noise
     data = (signal * snr) + (noise * (1 - snr))
--- a/mne_connectivity/datasets/frequency.py
+++ b/mne_connectivity/datasets/frequency.py
@@ -56,7 +56,8 @@ def make_signals_in_freq_bands(
         Names of the channels in the simulated data. If `None`, the channels are named
         according to their index and the frequency band of interaction.
     ch_types : str | list of str (default "eeg")
-        Types of the channels in the simulated data.
+        Types of the channels in the simulated data. If specified as a list, must be a
+        list of ``n_seeds + n_targets`` channel names.
     rng_seed : int | None (default None)
         Seed to use for the random number generator. If `None`, no seed is specified.
 
--- a/mne_connectivity/datasets/frequency.py
+++ b/mne_connectivity/datasets/frequency.py
@@ -1,0 +1,146 @@
+# Authors: Adam Li <adam2392@gmail.com>
+#          Thomas S. Binns <t.s.binns@outlook.com>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+from mne import EpochsArray, create_info
+from mne.filter import filter_data
+
+
+def make_signals_in_freq_bands(
+    n_seeds,
+    n_targets,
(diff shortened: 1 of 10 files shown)


larsoner:
> This could be done, but it would require a considerable amount of effort tweaking these thresholds which I can't really invest the time for at the moment

Can you just add a `# TODO: Use <whatever> for this after tweaking tolerances` so we don't forget that this is a possibility? Seems okay to leave it to future work

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+        Signal-to-noise ratio of the simulated data in the range [0, 1].
+    connection_delay : int (default 5)
+        Number of timepoints for the delay of connectivity between the seeds and
+        targets. If > 0, the target data is a delayed form of the seed data. If < 0, the
+        seed data is a delayed form of the target data.
+    tmin : int | float (default 0)
+        Earliest time of each epoch.
+    ch_names : list of str | None (default None)
+        Names of the channels in the simulated data. If `None`, the channels are named
+        according to their index and the frequency band of interaction.
```
adam2392, line 57:
```suggestion
        Names of the channels in the simulated data. If `None`, the channels are named
        according to their index and the frequency band of interaction. If specified, must be a list of
        ``n_seeds + n_targets`` channel names.
```

Is this correct?

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+    ch_types : str | list of str (default "eeg")
```
adam2392, line 58: Perhaps 

```
If specified as a list, must be a list of ``n_seeds + n_targets`` channel names.
```

tsbinns:
@larsoner @adam2392 Thanks both for the feedback!

I replaced the simulations where no values were being checked (e.g. for error catching parameters, saving & loading, etc...), and marked 5 tests where the simulations could be replaced with `make_signals_in_freq_bands` following some tweaks to the acceptable values.

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+        frequency.
+    n_epochs : int (default 10)
+        Number of epochs in the simulated data.
+    n_times : int (default 200)
+        Number of timepoints each epoch of the simulated data.
+    sfreq : int | float (default 100)
+        Sampling frequency of the simulated data, in Hz.
+    trans_bandwidth : int | float (default 1)
+        Transition bandwidth of the filter to apply to isolate activity in
+        ``freq_band``, in Hz.
```
adam2392, line 46:
```suggestion
    trans_bandwidth : int | float (default 1)
        Transition bandwidth of the filter to apply to isolate activity in
        ``freq_band``, in Hz. These are passed to the ``l_bandwidth`` and ``h_bandwidth``
        keyword arguments in :func:`mne.filter.create_filter`.
```

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+        sfreq=sfreq,
+        l_freq=freq_band[0],
+        h_freq=freq_band[1],
+        l_trans_bandwidth=trans_bandwidth,
+        h_trans_bandwidth=trans_bandwidth,
+        fir_design="firwin2",
+    )
+
+    # simulate noise for each channel
+    noise = rng.randn(n_channels, n_epochs * n_times + np.abs(connection_delay))
```
adam2392, line 118:
```suggestion
    noise = rng.standard_normal(size=(n_channels, n_epochs * n_times + np.abs(connection_delay)))
```

https://numpy.org/doc/stable/reference/random/generated/numpy.random.randn.html

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+    if snr < 0 or snr > 1:
+        raise ValueError("Signal-to-noise ratio must be between 0 and 1.")
+
+    if np.abs(connection_delay) >= n_epochs * n_times:
+        raise ValueError(
+            "Connection delay must be less than the total number of timepoints."
+        )
+
+    # simulate data
+    rng = np.random.RandomState(rng_seed)
```
adam2392, line 103:
```suggestion
    rng = np.random.default_rng(rng_seed)
```

mne_connectivity/datasets/frequency.py
```diff
@@ -0,0 +1,146 @@
...
+
+    # simulate signal source at desired frequency band
+    signal = rng.randn(1, n_epochs * n_times + np.abs(connection_delay))
```
adam2392, line 106:
```suggestion
    signal = rng.standard_normal(size=(1, n_epochs * n_times + np.abs(connection_delay)))
```

adam2392 (changes requested):
Some minor changes, but this overall looks good to me

mne_connectivity/datasets/frequency.py
tsbinns: Good to know!

tsbinns:
Looks like some of the thresholds were just on the boundary and updating the way in which the random numbers were generated caused these to fail, so tweaked the thresholds slightly.

adam2392 (approved):
Nice!

adam2392:
I'll let @larsoner and @drammock take a quick look.

Perhaps @ruuskas is also interested given he found a bug with the previous implementation? Hopefully this will simplify life for generating simulated data with expected correlations in the frequency domain

adam2392:
Perhaps something to help move this forward is refactoring #163 to use this function? (I.e. merge changes might work?...)

I would keep a local copy in case git messes up. 

Then we can see the simulation function work as intended for the new PR?

If that's a bit hairy, I'm okay merging for now. Wdyt @larsoner

tsbinns:
@adam2392 Sure, I'm open to this. Will be away at a conference rest of this week, but can start on Monday.

adam2392:
No rush! Enjoy!

tsbinns:
@adam2392 I've refactored #163 to use the new simulation function in the CaCoh and coherency method comparison examples where I was previously using a function within each example file.
Everything still works, no problems switching to the dedicated function.

adam2392:
Ah very cool! I think this LGTM as a start to simplifying data generation in our connectivity analyses.

Issue #174, opened 2024-03-06
Memory issues introduced in spectral_connectivity_time by #138


ruuskas:
#### Describe the bug
It appears that upon the addition of multivariate connectivity methods in #138, the following was introduced on line 483 in [mne_connectivity/spectral/time.py](https://github.com/tsbinns/mne-connectivity/blob/4c9548ca6cfb3f1d8f7a2b70ef30631f2daeda78/mne_connectivity/spectral/time.py). 
```
conn_patterns[m] = np.full((n_epochs, 2, len(source_idx), n_freqs),
                           np.nan)
```
and this was later changed to 
```
conn_patterns[m] = np.full(
    (n_epochs, 2, n_cons, max_n_channels, n_freqs), np.nan)
```
in #142.

This leads to memory problems with real data. In my current case, the array becomes of size
`(7, 2, 100128, 100128, 32)`, which would require 32 TiB of memory. This problem does not occur with version 0.5. 

#### Steps to reproduce
Try running the standard example with large number of signals (448 in my case) and frequencies.

#### Expected results
Code should run normally.

#### Actual results
I receive the following error:
```
File "<decorator-gen-290>", line 12, in spectral_connectivity_time
  File "*/miniconda3/envs/connectivity-pipeline/lib/python3.12/site-packages/mne_connectivity/spectral/time.py", line 543, in spectral_connectivity_time
    conn_patterns[m] = np.full(
                       ^^^^^^^^
  File "*/miniconda3/envs/connectivity-pipeline/lib/python3.12/site-packages/numpy/core/numeric.py", line 329, in full
    a = empty(shape, dtype, order)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 32.7 TiB for an array with shape (7, 2, 100128, 100128, 32) and data type float64
```

#### Additional information
I'm running MNE-Python v1.6 and MNE-Connectivity v0.6. 

Pinging here already @adam2392 and @tsbinns. For a quick fix, I guess the `conn_patterns` array could be initialized only if a multivariate method is used?

tsbinns:
Sorry for this @ruuskas, was an oversight on my part. I just submitted PR #175 to fix this by only initialising the array when it is required.

PR #175, opened 2024-03-06
[BUG] Fix memory error spectral_connectivity_time

tsbinns:
Fix for #174.

In short, the array for storing spatial patterns of connectivity used by the MIC method was unnecessarily being created for all methods. This could lead to memory errors when bivariate connectivity methods were being called on large numbers of connections.

This fix makes it so that the patterns array will only be instantiated in methods that require it, otherwise will be set to `None`. This does not cause any change in behaviour, since for methods where patterns are not used, `None` was already being stored in the `patterns` attribute of the returned connectivity objects.

This is a problem specific to `spectral_connectivity_time()`; `spectral_connectivity_epochs()` was already using the fixed approach.

### DIFF
--- a/mne_connectivity/spectral/epochs_multivariate.py
+++ b/mne_connectivity/spectral/epochs_multivariate.py
@@ -819,3 +819,4 @@ _CON_METHOD_MAP_MULTIVARIATE = {
 
 _multivariate_methods = ["mic", "mim", "gc", "gc_tr"]
 _gc_methods = ["gc", "gc_tr"]
+_patterns_methods = ["mic"]  # methods with spatial patterns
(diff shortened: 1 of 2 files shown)


adam2392 (commented):
Is there a unit-test that we can add to prevent this from occurring, or is this an internal detail?

tsbinns:
> Is there a unit-test that we can add to prevent this from occurring, or is this an internal detail?

Personally, I would consider this more of a one-off oversight in the implementation from a mistake on my end.

Tbh I wouldn't know where to start in defining an appropriate memory usage for a function such as this.

adam2392 (approved):
SG! LGTM minus a small nit

drammock (approved):

Issue #176, opened 2024-03-12
[BUG] n_jobs=-1 not working in spectral_connectivity_epochs


tsbinns:
#### Describe the bug
Setting `n_jobs=-1` for `spectral_connectivity_epochs()` fails to parallelise computation of the CSD over epochs.

If `n_jobs=-1`, `parallel_func()` is called to create an `_epoch_spectral_connectivity()` function associated with the appropriate number of workers from `multiprocessing.cpu_count()`.
https://github.com/mne-tools/mne-connectivity/blob/f9a2e658bcf05c6af61b937798a3b2e11e187599/mne_connectivity/spectral/epochs.py#L914-L917

However, when subsequently creating the epoch blocks to compute the CSD for in parallel, `_get_n_epochs()` does not convert `n_jobs=-1` to the maximum number of workers, which results in each epoch block only consisting of a single epoch, which are passed into the parallelised `_epoch_spectral_connectivity()` function one-by-one.
https://github.com/mne-tools/mne-connectivity/blob/f9a2e658bcf05c6af61b937798a3b2e11e187599/mne_connectivity/spectral/epochs.py#L991

#### Steps to reproduce
```
from multiprocessing import cpu_count
import numpy as np
from mne_connectivity import spectral_connectivity_epochs

n_epochs = 10
n_chans = 2
n_times = 100
sfreq = 50
data = np.random.rand(n_epochs, n_chans, n_times)

spectral_connectivity_epochs(data, sfreq=sfreq, n_jobs=cpu_count())
spectral_connectivity_epochs(data, sfreq=sfreq, n_jobs=-1)
```

#### Expected results
Both `n_jobs=cpu_count()` and `n_jobs=-1` should parallelise CSD computation over epochs.

#### Actual results
`n_jobs=cpu_count()` parallelised CSD computation over epochs as expected. E.g. with 16 workers on my machine I get the following logging output:
![image](https://github.com/mne-tools/mne-connectivity/assets/56922019/15dc4184-fb00-4dad-adf2-cdb2c545cf2f)

`n_jobs=-1` allocates multiple workers to `_epoch_spectral_connectivity()`, but only passes in the data of one epoch at a time. E.g. I get the following logging output:
![image](https://github.com/mne-tools/mne-connectivity/assets/56922019/00d5c90d-5e88-40bf-977f-c12d7f26c72d)

#### Bug fix
Fixing this is simply a case of having `parallel_func()` return `n_jobs` as the third output.
[...]

PR #177, opened 2024-03-12
[BUG] n_jobs=-1 not working in spectral_connectivity_epochs

tsbinns:
Suggested fix for #176 where `n_jobs=-1` does not result in CSD computation being properly parallelised over epochs.

This fix makes sure `n_jobs=-1` is converted to the maximum number of workers before CSD computation occurs.

I think also when refactoring some of the multivariate connectivity methods some months ago the `n_jobs` parameter was not being properly passed to the classes. This is also now fixed.

As far as I understand, these issues are specific to `spectral_connectivity_epochs()` and do not affect `spectral_connectivity_time()`.

The current unit tests for parallelisation were only using `n_jobs > 1` and never `n_jobs=-1`. However, even if `n_jobs=-1` was being used this issue would not be caught, as the tests only check that multiple workers can be allocated (which they can) but not that e.g. the proper number of epoch blocks are worked on in parallel.
Not sure how this could be checked. Is there some way the logging output could be queried?

### DIFF
--- a/mne_connectivity/spectral/epochs.py
+++ b/mne_connectivity/spectral/epochs.py
@@ -912,7 +912,7 @@ def spectral_connectivity_epochs(
     .. footbibliography::
     """
     if n_jobs != 1:
-        parallel, my_epoch_spectral_connectivity, _ = parallel_func(
+        parallel, my_epoch_spectral_connectivity, n_jobs = parallel_func(
             _epoch_spectral_connectivity, n_jobs, verbose=verbose
         )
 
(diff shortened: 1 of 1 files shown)


larsoner (approved):

tsbinns:
OSF seems to be pretty unstable right now which is causing `build_docs` to fail.

larsoner:
> Not sure how this could be checked. Is there some way the logging output could be queried?

Well you could use
```
with mne.utils.catch_logging(verbose="debug") as log:
    ...
log = log.getvalue()
assert "Got -1" not in log
assert "after requesting -1" in log
```
or some better check. See the message you should get in

https://github.com/mne-tools/mne-python/blob/90067893e330c1941c055be22bc5f442ad320ec3/mne/parallel.py#L122

But even this would pass on `main` since this is in `parallel_func` of MNE. To catch this bug at the MNE-Connectivity level you'd have to get some sort of logging message from mne-connectivity itself, so you'd have to add it.

larsoner:
I'm fine either way -- if it's only a few lines and is `logger.debug` then you can add it if you want. But this is also a pretty clear bugfix to me as is so I'm okay merging now if you want to move on to other stuff

tsbinns:
Sure, then I'm happy to merge as well.

larsoner:
Thanks @tsbinns !

PR #185, opened 2024-05-23
[MAINT] Add logic for handling AverageTFRArray in old MNE versions

tsbinns:
## Problem

Since MNE-Python v1.7 the API of `time_frequency.AverageTFR` has been changed. The old API from <= v1.6 is now available in the new class `time_frequency.AverageTFRArray`.

The example `cwt_sensor_connectivity.py` uses `AverageTFR` with the old API, causing documentation building to fail when using MNE-Python >= v1.7.

## Solution

One possibility is to check whether the new class `AverageTFRArray` is available and import this (for >= v1.7), otherwise use the old `AverageTFR` (for v <= 1.6):

https://github.com/mne-tools/mne-connectivity/blob/397d6d737a09ee976f8814c7ad90ded1ea9c8609/examples/cwt_sensor_connectivity.py#L19-L28

Also necessary to explicitly set the `nave` parameter, as the new `AverageTFRArray` does not support this as a positional argument:

https://github.com/mne-tools/mne-connectivity/blob/397d6d737a09ee976f8814c7ad90ded1ea9c8609/examples/cwt_sensor_connectivity.py#L100

Very open to other ideas for implementing such a logic. Cheers!


tsbinns:
Again some strange MacOS-specific runtime errors like in #184.

examples/cwt_sensor_connectivity.py
```diff
@@ -19,7 +20,12 @@
 from mne import io
 from mne_connectivity import spectral_connectivity_epochs, seed_target_indices
 from mne.datasets import sample
-from mne.time_frequency import AverageTFR
+
+# XXX: remove logic once support for mne<1.7 is dropped
+try:
+    from mne.time_frequency import AverageTFRArray as AverageTFR
+except ImportError:
+    from mne.time_frequency import AverageTFR
```
drammock, line 28: hmm, I'm not crazy about including logic like this in a public-facing example/tutorial. Is it enough to code it whatever way works for the CI docs build, and include a comment to the effect of "if you're on some other version of MNE, use the other class"?

tsbinns: Sure, I'm happy to take this approach as well.

tsbinns: How do you feel about this?

https://github.com/mne-tools/mne-connectivity/blob/12f23ac888c9d55adaa3aa020b0af575ceebb733/examples/cwt_sensor_connectivity.py#L95-L96

larsoner:
Looks like comments have been address so I'll mark for merge-when-green, thanks @tsbinns !

larsoner (approved):

PR #186, opened 2024-05-24
[MAINT] Fix codespell error

tsbinns:
New codespell version catches a spelling mistake that was previously being missed: "scenarious" which should be "scenarios" in the `compare_coherency_methods.py` example.


tsbinns:
Failure to build docs addresed in #185.

larsoner:
Thanks @tsbinns

PR #187, opened 2024-05-24
MAINT: Test on macOS arm64

larsoner:
`macos-latest` now means `macos-14` which is arm64. There is some issue with `np.linalg.det` it seems. This PR should fix them.


larsoner:
Okay simplified CIs also by removing the style check in favor of just improving pre-commit. I also got rid of `black` in favor of `ruff` similar to MNE-Python.

larsoner:
All green except docs which are fixed by #185 so in it goes!

Issue #182, opened 2024-05-22
[GSOC] Add a 'decoding' module to expand flexibility of multivariate methods


tsbinns:
### Background

@larsoner @wmvanvliet @drammock

A key limitation with the existing implementation of multivariate connectivity methods in `spectral_connectivity_epochs()` and `spectral_connectivity_time()` is the incompatibility with standard decoding paradigms, e.g. fitting filters to some training data and applying them to other data. Note that this applies only to those methods which use filters (CaCoh and MIC; not MIM or GC). Additionally, the low computational cost of this approach (avoiding repeated re-calculation of filters) would make it much more feasible for real-time analyses e.g. for BCIs.

Following the `decoding` module of MNE-Python, a `decoding` module in MNE-Connectivity can be created containing classes for each compatible connectivity method, i.e. a `CaCoh` class and a `MIC` class. These classes would support:
- fitting filters to epoched data to optimise connectivity (`fit()` method); would also store the corresponding patterns of connectivity.
- transforming data based on the fitted filters (`transform()` method).
- visualising the filters and patterns as topomaps (`plot_filters()` and `plot_patterns()` methods).

Both classes will have the same methods & parameters, and also work very similarly internally, allowing for a base class that handles most of the functionality. I envision the following structure (will denote points I think are especially worth discussing with ***bold italics***):

<br>

### Class initialisation

Similar to the classes in MNE-Python's decoding module, parameters could include:
- `Info` object containing key information for the epoched data that will be fit to/transformed.
- `indices` specifying the channels to treat as seeds and targets.
- `fmin` and `fmax` specifying the frequency band in which connectivity should be optimised.
  - like for the `decoding.SSD` class, I think fitting filters for only a single frequency band is best.
- `tmin` and `tmax` specifying the time range to use when computing the CSD; default the whole time period.
- `n_components` specifying the number of filters to fit; default as many components as the rank of the data that will be fit to (so the exact number will need to be determined when the `fit()` method is called).
- `rank` specifying any rank subspace projection to perform on the CSD for the seeds and targets; default projecting only to rank of the data that will be fit to (so again the exact number must be determined later).
- ***options for determining the CSD computation.***
  - here there could be the option of using a Fourier or multitaper approach, but I am unsure about Morlet wavelets since I don't think it makes much sense in this context to have time-resolved filters (also not supported in the `SSD` class).

[...]

Issue #181, opened 2024-04-21
Release v0.7


adam2392:
A lot of new features courtesy of @tsbinns.

We should aim to make a release. @tsbinns any interest in trying this out?

tsbinns:
@adam2392 Sorry for the delay, I was on holiday the last 2 weeks.

Yes, I would be very interested!

adam2392:
Would a v0.7 release of current code be better before merging in any GSoC changes? 

linking: https://github.com/mne-tools/mne-connectivity/issues/183

I'm okay either way I suppose. cc: @drammock

drammock:
seems worth cutting a release now --- not necessarily because of impending GSoC changes, but because it's been ~5 months and there are some useful changes and bugfixes on `main`.

tsbinns:
@adam2392 I'm happy to help out if the offer still stands!

adam2392:
I don't have a walk-thru of how to do this exactly for mne-connectivity, but mne-bids has a writeup on how to do this. We just don't have a few of their steps tho.

https://github.com/mne-tools/mne-bids/wiki/How-to-make-a-release

Do you want to take a look and see if you can sketch out a sequence of steps that makes sense to you? Once you do it once, you'll have the exp to do it on any package :)

adam2392:
Let us know if you have questions

tsbinns:
Sounds good, I will try and get this done ASAP.

Though it might be worth waiting until #184 is merged since this has some nice efficiency improvements and is unrelated to the GSoC work. Thoughts?

adam2392:
Sounds good to me!

tsbinns:
WIP in #195

drammock:
> @adam2392 One of the steps mentioned in the wiki is uploading to and checking install from test-pypi, but I don't have access. Do you need to do this before moving forward, or is it good enough that the installation in the circleci passes?

Been a while since I've done this so I just tried and:

- @adam2392 and @larsoner are the only maintainers on PyPI for mne_connectivity, so I can't help
- the process is: create an account on test.pypi.org (requires setting up 2FA)
- generate an API token through account settings
- run the `twine upload...` command and at the username prompt enter `__token__` and at the password prompt enter the API token.  https://test.pypi.org/help/#apitoken

larsoner:
Rather than do all that I'd rather modernize with trusted artifact publishing on release

https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#usage

[...]

PR #192, opened 2024-05-28
[GSOC] Store multivariate connectivity filters

tsbinns:
## Problem

One requirement of the decoding module we will introduce in the GSoC project is access to the multivariate filters, which are currently computed in the connectivity estimator class, but never stored. Discussed in more detail in #182.

I thought submitting this as a separate PR from the wider decoding module would help keep diff size small and make reviewing the changes easier.

## Solution

Add a `filters` attribute to the multivariate connectivity estimator classes similar to the existing `patterns` attribute.

However, since this will only be required in the case of the connectivity estimator being used in the decoding module (not with `spectral_connectivity_epochs/time()` functions), also add a `store_filters` flag which is by default `False`.
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L197-L202

When `store_filters=True`, the `filters` attribute will be set as a placeholder array where filters can be stored (again, similar to how `patterns` works), however when `False`, `filters` will remain as `None` to reduce memory usage.
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L220-L223

It would be sufficient to pass `store_filters` to the `_MultivariateCohEstBase` estimator class and have the `filters` attribute there, however I have passed `store_filters` to the parent `_EpochMeanMultivariateConEstBase` class and have the `filters` attribute here. This keeps things in line with the `patterns` attribute which is also currently only used for daughters of the `_MultivariateCohEstBase`, but this may not be the case in the future if other filter-based non-coherency methods are implemented, so this approach is more flexible and forward-looking.
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L103-L119

Other than that, it's a very simple case of just storing the filters when appropriate.
**MIC**
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L474-L476

**CaCoh**
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L672-L674

Since `store_filters` is by default `False`, no behaviour has been changed. Since this is also an internal feature there are no new unit tests to add at this stage (would all come when the decoding module is implemented proper).

## Other points

When implementing this, I also realised it was also possible to refactor the `reshape_results()` method in the multivariate connectivity estimator classes which were separately implemented for the coherency and Granger methods but doing the same thing, so that has been cleaned up.
https://github.com/tsbinns/mne-connectivity/blob/fc7e2a71924f7ead47b8024a10458014c0043526/mne_connectivity/spectral/epochs_multivariate.py#L182-L189

### DIFF
--- a/mne_connectivity/spectral/epochs_multivariate.py
+++ b/mne_connectivity/spectral/epochs_multivariate.py
@@ -105,14 +105,18 @@ class _EpochMeanMultivariateConEstBase(_AbstractConEstBase):
 
     n_steps = None
     patterns = None
+    filters = None
     con_scores_dtype = np.float64
 
-    def __init__(self, n_signals, n_cons, n_freqs, n_times, n_jobs=1):
+    def __init__(
+        self, n_signals, n_cons, n_freqs, n_times, n_jobs=1, store_filters=False
+    ):
         self.n_signals = n_signals
         self.n_cons = n_cons
         self.n_freqs = n_freqs
         self.n_times = n_times
         self.n_jobs = n_jobs
+        self.store_filters = store_filters
 
         # include time dimension, even when unused for indexing flexibility
         if n_times == 0:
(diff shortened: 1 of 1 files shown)


tsbinns:
Even if there is currently no functinal change, I guess it is worth holding off on merging this until v0.7 is released (#181)??

larsoner (approved):
Agreed let's cut 0.7 then merge

tsbinns:
Changes we added in #193

Issue #183, opened 2024-05-22
[GSOC] Expand multivariate methods to support >1 component


tsbinns:
### Background

@larsoner @wmvanvliet @drammock

A key limitation with the existing implementation of multivariate connectivity methods in `spectral_connectivity_epochs()` and `spectral_connectivity_time()` is the fact that for the CaCoh and MIC methods, the connectivity estimates for only the first, strongest component is returned (i.e. the component with maximised connectivity). However, the other, weaker components of connectivity can still contain relevant information, just as is the case for other eigendecomposition-based optimisation methods like CSP or SSD.

### New features

As part of GSoC, we proposed expanding support for the number of connectivity components which can be returned to more than just the first.

One option is to add an `n_components` argument to `spectral_connectivity_epochs()` and `spectral_connectivity_time()`, just like for the `CSP` and `SSD` classes of MNE-Python's `decoding` module. To ensure compatibility with the existing behaviour, this parameter could have a default value of `1` (or also `None`; not sure if there is convention in MNE that should be followed??). Otherwise, if `n_components` was specified to be >1:
- we should check that `n_components` does not exceed the minimum rank of the seeds and targets.
  - e.g. if the seed channels had a rank of 5 and the target channels a rank of 3, `n_components` must be <= 3; it does not make sense conceptually to expand the number of connectivity components beyond this minimal value, e.g. use seed filters no. 4 & 5 with target filter no. 3.
- when computing connectivity, the specified number of `n_components` should be taken from the filters (these filters are already computed in the eigendecompositon, just not being used).
- the patterns for each used filter should also be computed (currently only the patterns for the first filter are being computed).
- ***store the connectivity/pattern components in the returned connectivity object.***
  - what dimensions the connectivity results take: `[connections x components x frequencies (x times)]`? `[connections x frequencies (x times) x components]`?
  - do we create a new set of classes for storing results with a components dimension, or use the existing classes? Back when we were intially implementing the multivariate methods I recall @adam2392 suggesting the possibility of a connectivity class with an open-ended number of dimensions.

There are definitely some implementation details which will need to be discussed, particularly regarding storage of the results.

<br>

### Examples/tutorials

I am open to either adding a short section to the existing CaCoh and MIC examples which show how the `n_components` argument can be used and how the results for different components can be interpreted, or perhaps it makes sense to have a separate tutorial (they are already quite long). ***Very open to your feedback on this.***

<br>

### Unit tests

[...]

PR #193, opened 2024-05-29
[GSOC] Add decoding module

tsbinns:
As described in #182.

This adds support for a base class with `fit` and `transform` methods, as well as daughter classes for the CaCoh and MIC methods. The code takes advantage of the MNE-Python classes for computing the CSD and the existing MNE-Connectivity estimator classes for computing the filters and patterns.

As of now there are no unit tests. These will be added.

As of now there are no examples/tutorials. These will be added.

As of now there are no plotting methods. These will come in a separate PR.

As of now there is only support for a single component. This is described more in #183 and will come in a separate PR.

To slightly reduce the size of this PR's diff, I opened a separate one (#192) which adds support for storing the filters used for CaCoh and MIC.


mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,474 @@
...
+    filters_: Optional[tuple] = None
+    patterns_: Optional[tuple] = None
+
+    _indices: Optional[tuple] = None
+    _rank: Optional[tuple] = None
+    _conn_estimator: Optional[_EpochMeanMultivariateConEstBase] = None
+
+    @property
+    def indices(self):
+        """Get ``indices`` parameter in the input format."""
```
wmvanvliet, line 35: what is the "input format"? Perhaps we can find a better term for this.

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,474 @@
...
+        # instantiate connectivity estimator and add CSD information
+        self._conn_estimator = self._conn_estimator(
+            n_signals=X.shape[1],
+            n_cons=1,
+            n_freqs=1,
+            n_times=0,
+            n_jobs=self.n_jobs,
+            store_filters=True,
+        )
+        self._conn_estimator.accumulate(con_idx=np.arange(csd.shape[0]), csd_xy=csd)
```
wmvanvliet, line 205: here again we assume that the connectivity estimater operates on the CSD. Is there a way to refactor this so that `_AbstractDecompositionBase.fit` is a more general and `_AbstractCoherenceDecomposition.fit` actually uses the CSD?

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -446,20 +458,18 @@ def _compute_mic(self, E, C, seed_idcs, target_idcs, U_bar_aa, U_bar_bb, con_i):
         alpha = V_seeds[times[:, None], freqs, :, w_seeds.argmax(axis=2)]
         beta = V_targets[times[:, None], freqs, :, w_targets.argmax(axis=2)]
 
+        # Part of Eqs. 46 & 47; i.e. transform filters to channel space
+        alpha_Ubar = np.matmul(U_bar_aa, np.expand_dims(alpha, axis=3))
+        beta_Ubar = np.matmul(U_bar_bb, np.expand_dims(beta, axis=3))
```
wmvanvliet, line 463: Do we use `np.matmul` here instead of the `@` operator because of specific broadcasting rules?

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,474 @@
...
+        """
+        self._check_X(X, ndim=(2, 3))
+        if self.filters_ is None:
+            raise RuntimeError(
+                "no filters are available, please call the `fit` method first"
+            )
+
+        # transform seed and target data
+        X_seeds = self.filters_[0].T @ X[..., self.indices[0], :]
+        X_targets = self.filters_[1].T @ X[..., self.indices[1], :]
```
wmvanvliet, line 328: This assumes the decomposition is always linear and without a bias term. Technically the ABC class cannot know how the given `_conn_estimator` wants to perform the decomposition and intuitively I would think this should call `_conn_estimator.transform()`. Would something like that make sense in the larger design?

wmvanvliet (commented):
Great start and super neat code. There are some design choices here regarding what to implement at what layer of abstraction that raise some questions (see comments in code).

mne_connectivity/spectral/epochs_multivariate.py
tsbinns: This is actually something I wanted to get thoughts on.

Back when I first added these methods I used `np.matmul` instead of `@` because I didn't see `@` used anywhere else in MNE-Connectivity, so I wasn't sure if the explicit `np.matmul` was preferred. But I since realised that `@` is used all over the place in the main MNE package.

I feel `np.matmul` hinders readability somewhat so I wanted to open a PR and switch to `@`, but it just hasn't been a priority for me. Perhaps after v0.7 is out I can do this.

mne_connectivity/decoding/decomposition.py
tsbinns: I get what you mean that this transformation should be abstracted more. But in that case my first thought would be to handle this transformation in something like the `_AbstractCoherenceDecomposition` you describe above. I.e. a similar approach to how you suggest abstracting `fit` could be handled.

I just feel adding a `transform` method to the connectivity estimators would be overburdening these classes. Currently everything we are using them for (e.g. CSD rank subspace projection, filter and pattern computations, transforming filters and patterns back into channel space) is also essential for their role in the `spectral_connectivity_epochs/time` functions, but the transformation of new data would very niche to this use-case.

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,474 @@
...
+        if self.n_components is None:
+            self.n_components = np.min(self.rank)
+        elif self.n_components > np.min(self.rank):
+            raise ValueError(
+                "`n_components` is greater than the minimum rank of the data"
+            )
+
+    def _compute_csd(self, X):
+        """Compute the cross-spectral density of the input data."""
+        # XXX: fix csd returning [fmin +1 bin to fmax -1 bin] frequencies
```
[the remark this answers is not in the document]
tsbinns: First off I want to say I am not against refactoring this stuff, but just a few notes...

> The notion of using the CSD should not be in the abstract base class, because this is specific to methods operating on coherence.

I would disagree the CSD is specific to coherency-based methods. As I understand it, every method in `spectral_connectivity_epochs/time` relies on the CSD being computed. So e.g. in future if there was some method for decomposition of PLI-based connectivity, this would not fit the `_AbstractCoherenceDecomposition` name. Perhaps something like `_AbstractSpectralDecomposition` would be more appropriate?

mne_connectivity/decoding/decomposition.py
tsbinns: As I mentioned above I think it would be better to have something like `_AbstractSpectralDecomposition` than `_AbstractCoherenceDecomposition` where CSD-based methods could be handled.

In this case the base `fit` could cover checking the input data and its rank. The question is whether the connectivity estimator should also be instantiated in the base class. Would we expect all decomposition methods to have an equivalent connectivity estimator (i.e. something that could also be used with `spectral_connectivity_epochs/time`)?

Then like you say, computing CSD and adding it to the connectivity estimator (or whatever alternative there is) could be handled in the daughter class. Maybe a `_fit` method is called?

Coming back to the base class, the filters and patterns could then be computed and extracted. This should be okay to do in the base class as long as the connectivity estimator/alternative use the same method/attribute names, e.g. `compute_con()` for computing the filters and patterns which are stored as `.filters` and `.patterns` attributes.

mne_connectivity/decoding/decomposition.py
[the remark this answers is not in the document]
wmvanvliet: yes, good point, that would be a better name

mne_connectivity/decoding/decomposition.py
wmvanvliet: Ok, well-argued :)

mne_connectivity/spectral/epochs_multivariate.py
wmvanvliet: `@` is preferred. There may be some old instanced of `np.matmul` left. Feel free to change these if it doesn't break anything. I read somewhere that the broadcasting rules are a little different. In this case, we explicitly call `expand_dims` so I don't think broadcasting is going to be an issue.

mne_connectivity/decoding/decomposition.py
wmvanvliet: > I just feel adding a `transform` method to the connectivity estimators would be overburdening these classes.

I don't feel that at all (or maybe I misunderstand the situation). All scikit-learn estimators have both `fit` and `transform`.

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,532 @@
...
+# :func:`~mne_connectivity.spectral_connectivity_epochs` to visualise the connectivity
+# in the first 30 epochs. We also plot bivariate coherence to demonstrate the
+# signal-to-noise enhancements this multivariate approach offers. As expected, we see a
+# peak in connectivity at 15-20 Hz decomposed by the spatial filters.
+
+# %%
+
+# Connectivity profile of first 30 epochs (filters fit to these epochs)
+con_cacoh_first = spectral_connectivity_epochs(
+    epochs.get_data(item=np.arange(N_EPOCHS // 2)),
```
wmvanvliet, line 124: as an ignorant outsider: why does a function that is called `spectral_connectivity_epochs` take a numpy array as first argument and not an `mne.Epochs` object?

mne_connectivity/spectral/epochs_multivariate.py
tsbinns: Okay I can make this switch. Thankfully we have the regression tests which should pick up if this messes anything up.

mne_connectivity/decoding/decomposition.py
tsbinns: Okay, maybe we can discuss this further in tomorrow's meeting.

examples/decoding/cohy_decomposition.py
tsbinns: `spectral_connectivity_epochs` and `spectral_connectivity_time` both require epoched data, either as an `mne.Epochs` object or a 3D NumPy array.

Here I am passing in the array because I also want to give only the first 30 epochs, and using this item indexing with `.get_data()` is quicker than splitting the data into two separate `Epochs` objects for the first and last 30 epochs.

wmvanvliet: You can index `Epoch` objects, so to get the first 30 epochs, you simply do `epochs[:30]`, which is super quick.

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,532 @@
...
+########################################################################################
+# In addition to assessing the validity of the approach, we can also look at the time
+# taken to run the analysis. Below we present a scenario resembling an online sliding
+# window approach typical of a BCI system. We consider the first 30 epochs to be the
+# training data that the filters should be fit to, and the last 30 epochs to be the
+# windows of data that the filters should be applied to, transforming and computing the
+# connectivity of each window (epoch) of data sequentially.
+#
+# Doing so, we see that once the filters have been fit, it takes only a few milliseconds
+# to transform each window of data and compute its connectivity.
```
wmvanvliet, line 264: In addition to it being quicker, would the traditional argument of overfitting also be valid here? If you compute the filters on a single epoch, the filters will be super-tuned to that specific epoch. The data is simulated with coherence in the 15-20 Hz frequency band. With this in mind, you can see in the plot:

- filters fitted on the first 30 epochs and then applied to the final 30 epochs nicely capture the coherence in the 15-20 Hz band and not much outside of that frequency band. (proper cross-validation)
- filters fitted on 30 epochs and then applied to the same 30 epochs also capture some coherence outside of the 15-20 Hz band

I would imagine fitting the filters to a single epoch and then applying them to that same epoch will show even more coherence outside the 15-20 Hz band, even though there should not be any coherence there given our simulation.

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,532 @@
...
+    sfreq=epochs.info["sfreq"],
+)
+class_duration = time.time() - start
+
+ax = plt.subplot(111)
+ax.plot(
+    con_mic_class.freqs,
+    np.abs(con_mic_class.get_data()[0]),
+    color=plt.rcParams["axes.prop_cycle"].by_key()["color"][2],
+    label="Decomposition class",
```
wmvanvliet, line 425:
```suggestion
    label="ImCoh",
```

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,532 @@
...
+    con_mic_class.freqs,
+    np.abs(con_mic_class.get_data()[0]),
+    label="MIC (decomposition\nclass)",
+)
+ax.axvspan(FMIN, FMAX, color="grey", alpha=0.2, label="Fitted freq. band")
+ax.set_xlabel("Frequency (Hz)")
+ax.set_ylabel("Connectivity (A.U.)")
+ax.set_title("MIC")
+plt.legend()
+plt.show()
```
wmvanvliet, line 485: Why is there a difference between `MIC (standard function)` and `MIC (decomposition class)` at all?

wmvanvliet, line 485: Perhaps explain once more that the difference between `MIC (standard function)` and `MIC (decomposition class)` is that the letter will fit a single filter for the entire band, whereas the standard function will not. Hence the estimated coherence differs.

wmvanvliet: Well, you do that in the Limitations section right after this. Hrm. Perhaps it could be annotated in the plot somehow? I just got confused a little why there would be any difference at all between using the function or the decomposition class. A little reminder would be nice.

wmvanvliet:
I like the tutorial (cannot really call it an "example" anymore, but mne-connectivity does not have a distinction between examples and tutorials). It explained a lot.

examples/decoding/cohy_decomposition.py
tsbinns: Good point, I can maybe introduce this earlier.

examples/decoding/cohy_decomposition.py
tsbinns: Quite right, I always forget these alternative indexing options. Have changed.

examples/decoding/cohy_decomposition.py
tsbinns: > In addition to it being quicker, would the traditional argument of overfitting also be valid here? If you compute the filters on a single epoch, the filters will be super-tuned to that specific epoch.

Yeah, that's a very good point. Will add this.

examples/decoding/cohy_decomposition.py
tsbinns: Have changed to match later figure.

tsbinns:
**FYI:** Added a set of unit tests. Coverage missing for two statements related to `n_components`, but they can't be added until support for >1 component supproted.

tsbinns:
@wmvanvliet I updated the way CSD parameters are handled to better support the `cwt_morlet` mode as we discussed yday: https://github.com/mne-tools/mne-connectivity/pull/193/commits/025e6c17cf1cbf00a54b5cc7054ee48943cbcaa1

tsbinns:
@wmvanvliet Refactored the decomposition classes to avoid over-future proofing in https://github.com/mne-tools/mne-connectivity/pull/193/commits/7b98dda99e9017ea8d8963567d43d1040c5729e8 as discussed yday.

Now everything is handled by a `CoherencyDecomposition` class which has a `method` parameter for controlling whether CaCoh or MIC is used.

mne_connectivity/decoding/tests/test_decomposition.py
```diff
@@ -0,0 +1,672 @@
...
+
+    # TEST BAD FITTING
+    # Test input data
+    with pytest.raises(TypeError, match="`X` must be an instance of NumPy array"):
+        decomp_class.fit(X=epochs.get_data().tolist())
+    with pytest.raises(ValueError, match="Invalid value for the '`X.ndim`' parameter"):
+        decomp_class.fit(X=epochs.get_data()[0])
+    with pytest.raises(ValueError, match="`X` does not match Info"):
+        decomp_class.fit(X=epochs.get_data()[:, :-1])
+    # XXX: Add test for rank of X being <= n_components when n_components can be > 1
```
wmvanvliet, line 655: You can quickly fake a rank-deficient `X` by doing a PCA to remove ranks:

```
from sklearn.decomposition import PCA
X = epochs.get_data().reshape(len(epochs), -1).T
p = PCA(n_components=10).fit(1)  # some n_components < n_channels
X = p.transform(X)
X = p.inverse_transform(X)
X = X.T.reshape(epochs.get_data().shape)
```

wmvanvliet:
Just one final `XXX` left and then this is good to merge from my end!

mne_connectivity/decoding/tests/test_decomposition.py
tsbinns: This is along the lines of how I would test this, however for now whilst `n_components` can only be 1 I am manually setting this:
https://github.com/mne-tools/mne-connectivity/blob/bdfb626382a8172fb5afad2c79ddafeb0e0e47fb/mne_connectivity/decoding/decomposition.py#L223

So even with rank-deficient data, the rank will always be >= `n_components`.

In the PR when support for >1 components is added then I would be able to clean up a few of these outstanding tests/missing coverage.

tsbinns: Updated to manually set `n_components` so these tests can be run.

tsbinns:
@wmvanvliet Added the extra test so pending CI checks (and v0.7 going out) this should be everything.

mne_connectivity/decoding/tests/test_decomposition.py
```diff
@@ -0,0 +1,691 @@
...
+            match=(
+                "`fmin` and `fmax` must not be None if `mode` is 'multitaper' or "
+                "'fourier'"
+            ),
+        ):
+            CoherencyDecomposition(
+                info=epochs.info,
+                method=method,
+                indices=indices,
+                mode=mode,
```
larsoner, line 341: If almost all of these take these four arguments perhaps it becomes more readable to do:
```
                **base_kwargs
```
and earlier have
```
base_kwargs = dict(
    info=epochs.info,
    method=method,
    indices=indices,
    mode=mode,
)
```
And as a nice side effect hopefully a lot of these calls can become one-liners rather than taking 5+ lines each.

mne_connectivity/spectral/epochs_multivariate.py
larsoner: looks like these and a few other places still need to be switched to `@`

larsoner (commented):
Just a couple of minor comments from me!

mne_connectivity/spectral/epochs_multivariate.py
tsbinns: Made the switch in a separate branch to make reviewing easier but didn't submit a PR for (didn't want to confuse things for the 0.7 release).

Have submitted now in #201. Regression tests pass, so no broadcasting silliness introduced.

tsbinns: Merged now

tsbinns:
> looks like these and a few other places still need to be switched to @

Now merged from #201

mne_connectivity/decoding/tests/test_decomposition.py
tsbinns: Good idea. Addressed in 88eafb5de21d2aba90cd23d7ad913888755a6332 & looks much nicer!

tsbinns:
That's comments of @wmvanvliet and @larsoner addressed!

larsoner (approved):
@wmvanvliet feel free to merge if you're happy!

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,472 @@
...
+                "`n_components` is greater than the minimum rank of the data"
+            )
+
+        # set rank if necessary
+        if self._rank is None:
+            self._rank = rank
+
+    def _compute_csd(self, X):
+        """Compute the cross-spectral density of the input data."""
+        # XXX: fix csd returning [fmin +1 bin to fmax -1 bin] frequencies
```
drammock, line 344: is this `XXX` comment meant to be addressed in this PR?

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,472 @@
...
+    def fit_transform(self, X, y=None, **fit_params):
+        """Fit filters to data, then transform and return it.
+
+        Parameters
+        ----------
+        X : array, shape=(n_epochs, n_signals, n_times)
+            The input data which the connectivity decomposition filters should be fit to
+            and subsequently transformed.
+        y : None
+            Used for scikit-learn compatibility.
```
drammock, line 440: [nitpick] can this be a bit clearer? I'm assuming the param needs to be in the signature for this class to be usable in sklearn pipelines; if that's right maybe say: "Ignored; exists for compatibility with scikit-learn pipelines" ?

drammock:
please wait, I am in the middle of reviewing now

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,557 @@
...
+# ``spectral_connectivity_...()`` functions: the transformed data; the indices
+# returned from
+# :meth:`~mne_connectivity.decoding.CoherencyDecomposition.get_transformed_indices`; and
+# the corresponding bivariate method (``"coh"`` and ``"cohy"`` for CaCoh; ``"imcoh"``
+# for MIC).
+#
+# As you can see, the connectivity profile of the transformed data using filters fit on
+# the first 30 epochs is very similar to the connectivity profile when using filters fit
+# on the last 30 epochs. This shows that the filters are generalisable, able to extract
+# the same components of connectivity which they were trained on from new data.
```
drammock, line 208: is it worth pointing out that connectivity estimates *outside* the 15-20 Hz band are *lower* when the filters are trained on different data (i.e., better SNR)?

drammock: I now see that @wmvanvliet had a similar comment to mine, in a resolved comment below. I do think though that it's worth mentioning the SNR point here too, in addition to the overfitting point made later using the sliding window example.

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,557 @@
...
+    fmax=30,
+    sfreq=epochs.info["sfreq"],
+)
+class_duration = time.time() - start
+
+ax = plt.subplot(111)
+ax.plot(
+    con_mic_class.freqs,
+    np.abs(con_mic_class.get_data()[0]),
+    color=plt.rcParams["axes.prop_cycle"].by_key()["color"][2],
```
drammock, line 452:
```suggestion
    color="C2",
```
this is shorthand equivalent for "the third color in the cycle" (works for `"C0"` through `"C9"`)

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,557 @@
...
+########################################################################################
+# There are two equivalent options for fitting and transforming the same data: 1)
+# passing the data to the :meth:`~mne_connectivity.decoding.CoherencyDecomposition.fit`
+# and :meth:`~mne_connectivity.decoding.CoherencyDecomposition.transform` methods
+# sequentially; or 2) using the combined
+# :meth:`~mne_connectivity.decoding.CoherencyDecomposition.fit_transform` method.
+#
+# We use the latter approach below, fitting the filters to the 15-20 Hz band and using
+# the ``"imcoh"`` method in the call to the ``spectral_connectivity_...()`` functions.
+# Plotting the results, we see a peak in connectivity at 15-20 Hz.
```
drammock, line 421: something is confusing me here. The first example faked some data with simulated connectivity in the 15-20Hz band, fit filters from 5-35 Hz, and recovered the fact that connectivity was higher in the 15-20Hz band.  But in this example, you load *real data* and fit filters on the 15-20Hz band, and transform the data using those filters... so IIUC the plot for case 2 basically just shows that the filter worked.  Do I have that right?  

If so, what I find confusing about it is how similar the plots are styled, as if they're showing basically the same thing on fake vs real data. But case 1 is showing "we can recover actual (simulated) connectivity and we don't see spurious connectivity where it doesn't exist", while case 2 is showing "the spatial filters actually capture connectivity in the band we asked for, and (mostly) suppress connectivity in frequencies outside that band".  I'm not totally sure how best to improve the tutorial to overcome this confusion; maybe first you can confirm that I'm actually understanding / interpreting correctly, then we can talk about how to improve it.

drammock (commented):
NB: I didn't review the tests.

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -189,9 +202,11 @@ class _MultivariateCohEstBase(_EpochMeanMultivariateConEstBase):
     name: Optional[str] = None
     accumulate_psd = False
 
-    def __init__(self, n_signals, n_cons, n_freqs, n_times, n_jobs=1):
+    def __init__(
+        self, n_signals, n_cons, n_freqs, n_times, n_jobs=1, store_filters=False
+    ):
```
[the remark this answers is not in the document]
wmvanvliet: I agree with this.

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -189,9 +202,11 @@ class _MultivariateCohEstBase(_EpochMeanMultivariateConEstBase):
...
         super(_MultivariateCohEstBase, self).__init__(
```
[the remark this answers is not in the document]
wmvanvliet: When a class only has a single superclass, I think just `super()` suffices (this didn't use to be the case in older Python versions)

mne_connectivity/decoding/decomposition.py
tsbinns: This has been solved, just forgot to remove. Thanks for pointing it out!

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -174,6 +178,15 @@ def reshape_csd(self):
             self._acc, (self.n_signals, self.n_signals, self.n_freqs, self.n_times)
         ).transpose(3, 2, 0, 1)
 
+    def reshape_results(self):
+        """Remove time dimension from results, if necessary."""
+        if self.n_times == 0:
+            self.con_scores = self.con_scores[..., 0]
```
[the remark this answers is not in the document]
tsbinns: Good idea, will make the switch.

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,472 @@
...
+                raise ValueError("`rank` must have length 2")
+            for rank_group in rank:
+                _validate_type(rank_group, int, "`rank`", "tuple of ints or None")
+        _rank = self._check_rank(rank, indices)
+
+        _validate_type(n_jobs, int, "`n_jobs`", "int")
+
+        _validate_type(
+            verbose, (bool, str, int, None), "`verbose`", "bool, str, int, or None"
+        )
```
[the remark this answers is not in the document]
tsbinns: You're right, will remove.

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,472 @@
...
+    def _check_X(self, X, ndim):
+        """Check that the input data is valid."""
+        # check data is a 2/3D array
+        _validate_type(X, np.ndarray, "`X`", "NumPy array")
+        _check_option("`X.ndim`", X.ndim, ndim)
+        n_chans = X.shape[-2]
+        if n_chans != self.info["nchan"]:
+            raise ValueError(
+                "`X` does not match Info\nExpected %i channels, got %i"
+                % (n_chans, self.info["nchan"])
```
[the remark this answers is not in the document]
tsbinns: Changed.

This old format is used a lot in MNE-Connectivity. Would people be open to a maint PR to switch to f-strings?

mne_connectivity/spectral/epochs_multivariate.py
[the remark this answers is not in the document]
tsbinns: Sounds good, will change.

mne_connectivity/spectral/epochs_multivariate.py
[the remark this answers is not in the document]
drammock: In Python 3, within a class definition it will always work to use the zero-argument form of `suoer()`, even in cases of multi inheritance.

mne_connectivity/decoding/decomposition.py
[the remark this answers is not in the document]
drammock: Yes!

mne_connectivity/spectral/epochs_multivariate.py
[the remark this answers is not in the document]
tsbinns: Yeah, this was a holdover from the previous connectivity classes. Happy to address this in a maint PR so will open an issue as a reminder.

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -210,6 +225,10 @@ def compute_con(self, indices, ranks, n_epochs=1):
             self.patterns = np.full(
                 (2, self.n_cons, indices[0].shape[1], self.n_freqs, n_times), np.nan
             )
+            if self.store_filters:
+                self.filters = np.full(
+                    (2, self.n_cons, indices[0].shape[1], self.n_freqs, n_times), np.nan
```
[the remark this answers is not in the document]
tsbinns: Agreed!

mne_connectivity/decoding/decomposition.py
```diff
@@ -0,0 +1,472 @@
...
+                )
+                _validate_type(mt_adaptive, bool, "`mt_adaptive`", "bool")
+                _validate_type(mt_low_bias, bool, "`mt_low_bias`", "bool")
+        else:
+            if cwt_freqs is None:
+                raise TypeError(
+                    "`cwt_freqs` must not be None if `mode` is 'cwt_morlet'"
+                )
+            _validate_type(
+                cwt_freqs, (tuple, list, np.ndarray), "`cwt_freqs`", "array-like"
```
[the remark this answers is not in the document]
tsbinns: Thanks for pointing out! There are a few places I can tidy this up.

mne_connectivity/decoding/decomposition.py
drammock: Yeah, I think it's worth propagating that docstring change upstream

examples/decoding/cohy_decomposition.py
tsbinns: Can't believe I didn't know this...

mne_connectivity/spectral/epochs_multivariate.py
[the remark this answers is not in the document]
tsbinns: > I'm guessing it's elsewhere in the codebase and should be fixed everywhere

Yeah, this was a holdover of previous connectivity classes. Happy to address in a maint PR, will make an issue as a reminder.

examples/decoding/cohy_decomposition.py
tsbinns: Can't believe I didn't know of this... Thanks!

tsbinns:
Thanks @drammock for the detailed feedback!

I have addressed everything except for the bigger points about the example's clarity. Will try to work on those now.

mne_connectivity/decoding/decomposition.py
tsbinns: Okay, submitted a PR: https://github.com/mne-tools/mne-python/pull/12664

examples/decoding/cohy_decomposition.py
tsbinns: Sure, I can add this. How about something like:

"As you can see, the connectivity profile of the transformed data using filters fit on the first 30 epochs is very similar to the connectivity profile when using filters fit on the last 30 epochs. This shows that the filters are generalisable, able to extract the same components of connectivity which they were trained on from new data. Additionally, by fitting the filters to only a select frequency band, we can avoid overfitting in frequency bins where no interactions are present (here outside of the 15-20 Hz band) and thereby enhance the overall signal-to-noise ratio of our connectivity estimates."

examples/decoding/cohy_decomposition.py
tsbinns: > case 1 is showing "we can recover actual (simulated) connectivity and we don't see spurious connectivity where it doesn't exist", while case 2 is showing "the spatial filters actually capture connectivity in the band we asked for, and (mostly) suppress connectivity in frequencies outside that band"

So case 1 is just about showing that you can fit filters to one piece of data and apply it to another (not necessarily about showing there is no spurious connectivity), and case 2 is about fitting filters to and transforming the same piece of data (not necessarily that the filters are capturing connectivity in a given band and suppressing it elsewhere). For either/both of these I could have used real/simulated data.

For case 1 I looked at using the fieldtrip_cmc data for this, but found it difficult to identify a connectivity component that was present across multiple parts of the recording duration (which is necessary if you want to fit the filters to one set of epochs and apply them to another). I'm sure I could have explored this more, but in the interest of time I settled on using simulated data where I know there is a consistent connectivity component throughout the whole length of data.

For case 2 I could have used simulated data again, but I always think these tutorials look more convincing with some real data examples, and for this case since we are fitting to and transforming the same data, identifying a connectivity component which is consistent across multiple parts of the recording is not necessary.

<br>

> The first example faked some data with simulated connectivity in the 15-20Hz band, fit filters from 5-35 Hz, and recovered the fact that connectivity was higher in the 15-20Hz band

For case 1, I start by showing the connectivity of the data that our simulated data indeed involves an interaction at 15-20 Hz. I compute connectivity using the `spectral_connectivity_epochs()` function where filters are fit to each frequency bin.

I then use the new decomposition class to fit filters to **only the 15-20 Hz range** which are trained on the first 30 epochs and applied to the last 30 epochs. Alongside this I show the results of the `spectral_connectivity_epochs()` function for the last 30 epochs (i.e. fitting and transforming these epochs, with filters fit to each frequency bin).

**This is to demonstrate that the filters trained on one piece of data can extract this same connectivity component from a new piece of data**. The point is not to show that the filters do not pick up spurious connectivity.

<br>

> you load real data and fit filters on the 15-20Hz band, and transform the data using those filters... so IIUC the plot for case 2 basically just shows that the filter worked.

In case 2 I show that you can also use the decomposition class to fit and transform the same piece of data, similar to what is happening in the `spectral_connectivity_...` functions (except with fitting filters to a single band, not each frequency bin).

In the same way as for case 1 I am still only fitting filters to 15-20 Hz using the decomposition class. The first plot where the results of fitting & transforming the same data using the decomposition classes is shown alone is just to demonstrate that there is connectivity at 15-20 Hz in the data.

Then I compute connectivity using the `spectral_connectivity_epochs()` function (where filters are fit to each frequency bin) to show that fitting and transforming the same piece of data using the decomposition class gives a near-identical result to the `spectral_connectivity_...` approach, at least for the 15-20 Hz range where the decomposition class fit the filters.

**The point is to demonstrate that fitting to and transforming the same piece of data with the decomposition class is a valid use-case**. The point is not that connectivity outside the given frequency band is suppressed (it's not actively suppressed, the filters just don't care about optimising this connectivity). It does show that "the spatial filters actually capture connectivity in the band we asked for", but this is also shown in case 1.

<br>

Does this make it any clearer about what each example is trying to show? These are the key points:
- case 1 is just about showing that you can fit filters to one piece of data and apply it to another
- case 2 is about fitting filters to and transforming the same piece of data

Maybe having a final summary after case 1 & 2 helps get this across.

wmvanvliet: > This is to demonstrate that the filters trained on one piece of data can extract this same connectivity component from a new piece of data. The point is not to show that the filters do not pick up spurious connectivity.

But perhaps the tutorial can be changed to include the point that the filters do not pick up spurious connectivity. This is an important argument in support of the statement "can extract the same connectivity component".

At any rate, I feel this is something we want users to think about when they make decisions regarding what parts of the signal to fit the filters on and what parts to apply the filters to. The tutorial can help set them on the right path.

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,557 @@
...
+########################################################################################
+# Limitations
+# -----------
+# Finally, it is important to discuss a key limitation of the decoding module approach:
+# the need to define a specific frequency band. Defining this band requires some
+# existing knowledge about your data or the oscillatory activity you are studying. This
+# insight may come from a pilot study where a frequency band of interest was identified,
+# a canonical frequency band defined in the literature, etc... In contrast, by fitting
+# filters to each frequency bin, the standard ``spectral_connectivity_...()`` functions
+# are more flexible.
```
drammock, line 538:
```suggestion
# are more data-driven.
```

examples/decoding/cohy_decomposition.py
```diff
@@ -0,0 +1,557 @@
...
+#
+# Additionally, by applying filters fit on one set of data to another, you are assuming
+# that the connectivity components the filters are designed to extract are consistent
+# across the two sets of data. However, this may not be the case if you are applying the
+# filters to data from a distinct functional state where the spatial distribution of the
+# components differs. Again, by fitting filters to each new set of data passed in, the
+# standard ``spectral_connectivity_...()`` functions are more flexible, extracting
```
drammock, line 545:
```suggestion
# standard ``spectral_connectivity_...()`` functions are more data-driven, extracting
```

drammock (commented):
> I always think these tutorials look more convincing with some real data examples

I generally agree, as long as the tutorial says things like "from the literature, we expect to find XYZ [cite,cite] reflecting the brain doing ABC" (so, e.g., "connectivity in beta band reflecting a cortico-cerebellar loop for motor planning" or whatever), and there's a dataset available that shows the effect.

> case 1 is just about showing that you can fit filters to one piece of data and apply it to another (not necessarily about showing there is no spurious connectivity), and case 2 is about fitting filters to and transforming the same piece of data

I think if the same data were used for both of these, the tutorial could be easier to follow. Something like:

1. simulate data
2. fit the filters to first half of epochs
3. apply to same epochs & visualize result (& compare with `spect_conn_epo` function)
4. now apply to held-out epochs & visualize result

The narrative could be something like "Case 1: the new Decomp Class can do what the existing function did (fit/transform to same data). Case 2: the new Decomp Class can *also* be used in this other way (fit/transform on separate train/test data)."

> it's not actively suppressed, the filters just don't care about optimising this connectivity

Fair enough, "suppressed" was not a good word choice on my part. I think it's relevant to discuss the out-of-band frequencies though.  E.g., you could say things like "here we don't see much out-of-band connectivity, but in theory a filter trained on 15-25Hz might also pick up connectivity in other bands, if the spatial pattern of the connectivity in those other bands happened to coincide with the spatial pattern that maximizes our band of interest."

tsbinns:
@drammock

> I think if the same data were used for both of these, the tutorial could be easier to follow.

Okay, then I will switch to simulated data for both.

> I think it's relevant to discuss the out-of-band frequencies though.

Sure, and that also align's with @wmvanvliet's [comment](https://github.com/mne-tools/mne-connectivity/pull/193#discussion_r1642230208). I will add a section on this as well.

examples/decoding/cohy_decomposition.py
tsbinns: Added

tsbinns:
Have restructured the example/tutorial based on people's comment. Thanks again @drammock & @wmvanvliet!

For the point about how activity outside of the fitting freq. range is handled, I just added text explaining this. Could also add some examples showing this, but I would only do that if people think it is necessary.

https://github.com/mne-tools/mne-connectivity/blob/29f2b53c51dbacfb7662e33ac69f4cc41333a23c/examples/decoding/cohy_decomposition.py#L493-L523

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
+# ``spectral_connectivity_...()`` functions.
+#
+# To compute connectivity of the transformed data, it is simply a case of passing to the
+# ``spectral_connectivity_...()`` functions: the transformed data; the indices returned
+# from
+# :meth:`~mne_connectivity.decoding.CoherencyDecomposition.get_transformed_indices`; and
+# the corresponding bivariate method (``"coh"`` and ``"cohy"`` for CaCoh; ``"imcoh"``
+# for MIC).
+#
+# For reference, we will also compute connectivity using the standard CaCoh approach of
```
drammock, line 167:
```suggestion
# For comparison, we will also compute connectivity using the standard CaCoh approach of
```

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
+# the ``spectral_connectivity_...()`` functions, as well as bivariate coherence to show
+# the signal-to-noise enhancements the multivariate approach offers.
```
drammock, line 169:
```suggestion
# the signal-to-noise benefits of the multivariate approach.
```

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
+#
+# To compute connectivity of the transformed data, it is simply a case of passing to the
+# ``spectral_connectivity_...()`` functions: the transformed data; the indices returned
+# from
+# :meth:`~mne_connectivity.decoding.CoherencyDecomposition.get_transformed_indices`; and
+# the corresponding bivariate method (``"coh"`` and ``"cohy"`` for CaCoh; ``"imcoh"``
+# for MIC).
+#
+# For reference, we will also compute connectivity using the standard CaCoh approach of
+# the ``spectral_connectivity_...()`` functions, as well as bivariate coherence to show
```
drammock, line 168:
```suggestion
# the ``spectral_connectivity_...()`` functions, as well as bivariate coherence, to show
```

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
-# ``spectral_connectivity_...()`` functions: the transformed data; the indices
-# returned from
-# :meth:`~mne_connectivity.decoding.CoherencyDecomposition.get_transformed_indices`; and
-# the corresponding bivariate method (``"coh"`` and ``"cohy"`` for CaCoh; ``"imcoh"``
-# for MIC).
+# Case 2: Fitting to and transforming different data
+# --------------------------------------------------
+# Another way we can use the decomposition class is by taking the filters trained on one
+# piece and applying them to another piece of data. Continuing with our simulated data
+# example, we will use the filters fit on the first 30 epochs of data.
```
drammock, line 276:
```suggestion
# example, we can re-use the :class:`~mne_connectivity.decoding.CoherencyDecomposition`
# instance we made earlier, as it already contains filters fit on the first 30 epochs of
# data.
```

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
 #
-# As you can see, the connectivity profile of the transformed data using filters fit on
-# the first 30 epochs is very similar to the connectivity profile when using filters fit
-# on the last 30 epochs. This shows that the filters are generalisable, able to extract
-# the same components of connectivity which they were trained on from new data.
+# Using the :meth:`~mne_connectivity.decoding.CoherencyDecomposition.transform` method,
+# we apply the filters from the first 30 epochs to the last 30 epochs of data,
```
drammock, line 279:
```suggestion
# We can then apply those filters to the last 30 epochs of data,
```

(also I suggest merging this sentence with the prior paragraph)

examples/decoding/cohy_decomposition.py
```diff
@@ -108,173 +109,266 @@
...
-# windows of data that the filters should be applied to, transforming and computing the
-# connectivity of each window (epoch) of data sequentially.
-#
-# Doing so, we see that once the filters have been fit, it takes only a few milliseconds
-# to transform each window of data and compute its connectivity.
+# Again, notice how the CaCoh results from the decomposition class show less
+# connectivity outside of the 15-20 Hz range compared to the CaCoh results of the
+# ``spectral_connectivity_...()`` functions.
+
+########################################################################################
```
drammock, line 357:
```suggestion
#
```

drammock (approved):
Really nice improvements. Just a few minor nitpicks / wordsmithings. (also it looks like you have a merge conflict to resolve)

wmvanvliet:
Thanks heaps @tsbinns !

PR #198, opened 2024-06-04
[MAINT] Add release workflow

tsbinns:
@larsoner suggested in #181 to modernise publishing to PyPI with trusted artefact publishing on release.

Was suggested to use the MNE-BIDS-Pipeline workflow as a template and removing the username/password stuff: https://github.com/mne-tools/mne-bids-pipeline/blob/main/.github/workflows/release.yml

Haven't touched the existing `publish_release.yml` workflow. Very minor but wasn't sure if we should use this name, rename to `release.yml`, ygm.

Also had to add a new rule to the `.yamllint.yml` config for pre-commit yamllint check to pass on my windows machine due to unix vs. dos line endings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,51 @@
+name: Build
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  package:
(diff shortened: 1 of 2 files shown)


larsoner (approved):
I'll merge but we'll need to wait for @adam2392 to make me an owner on PyPI, then we can set up trusted publishing and try it!

Thanks @tsbinns !

PR #205, opened 2024-06-18
[MAINT] Clean up string formatting

tsbinns:
Does a few things:

1. Switches to f-string formatting, hopefully didn't miss any (requested by @drammock: https://github.com/mne-tools/mne-connectivity/pull/193#discussion_r1639558215)
2. Gets rid of annoying leftovers from switch to 88 line length where strings that were previously split over two lines no longer are, but were kept as two separate strings, e.g.:
    ```
    "this used to be split over two" "lines" => "this used to be split over two lines"
    ```
3. Takes advantage of the new 88 line length for strings to reduce code footprint

(2) & (3) weren't requested, I just thought they would be good. If people would rather stick to just (1) I can change this.

### DIFF
--- a/mne_connectivity/base.py
+++ b/mne_connectivity/base.py
@@ -64,10 +64,7 @@ class EpochMixin:
         if events is not None:
             for key, val in self.event_id.items():
                 if val not in events[:, 2]:
-                    msg = "No matching events found for %s " "(event id %i)" % (
-                        key,
-                        val,
-                    )
+                    msg = f"No matching events found for {key} (event id {val})"
                     _on_missing(on_missing, msg)
 
             # ensure metadata matches original events size
@@ -95,9 +92,8 @@ class EpochMixin:
         """
         if not isinstance(self, type(epoch_conn)):
             raise ValueError(
-                f"The type of the epoch connectivity to append "
-                f"is {type(epoch_conn)}, which does not match "
-                f"{type(self)}."
+                f"The type of the epoch connectivity to append is {type(epoch_conn)}, "
+                f"which does not match {type(self)}."
             )
         if hasattr(self, "times"):
             if not np.allclose(self.times, epoch_conn.times):
@@ -157,8 +153,8 @@ class EpochMixin:
 
         if not self.is_epoched:
             raise RuntimeError(
-                "Combine only works over Epoched connectivity. "
-                f"It does not work with {self}"
+                "Combine only works over Epoched connectivity. It does not work with "
+                f"{self}"
             )
 
         fun = _check_combine(combine, valid=("mean", "median"))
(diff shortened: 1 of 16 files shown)


tsbinns:
Just realised I missed some things. Will fix now.

mne_connectivity/base.py
```diff
@@ -419,10 +415,9 @@ def __init__(
     ):
         if isinstance(indices, str) and indices not in ["all", "symmetric"]:
             raise ValueError(
-                f"Indices can only be "
-                f'"all", otherwise '
-                f"should be a list of tuples. "
-                f"It cannot be {indices}."
+                "Indices can only be "
+                '"all", otherwise '
+                f"should be a list of tuples. It cannot be {indices}."
```
drammock, line 420:
```suggestion
                'Indices can only be "all", otherwise '
                f"should be a list of tuples. It cannot be {indices}."
```

also wondering why `"symmetric"` isn't mentioned in the Error as a valid option?

mne_connectivity/tests/test_connectivity.py
```diff
@@ -144,9 +144,7 @@ def test_connectivity_containers(conn_cls):
     conn = conn_cls(data=correct_numpy_input, n_nodes=3, **extra_kwargs)
 
     # test that get_data works as intended
-    with pytest.raises(
-        ValueError, match="Invalid value for the " "'output' parameter*."
-    ):
+    with pytest.raises(ValueError, match="Invalid value for the 'output' parameter*."):
```
drammock, line 147:
```suggestion
    with pytest.raises(ValueError, match="Invalid value for the 'output' parameter"):
```
probably meant to be `.*` but explicit `.*` is not required at the end of the match string

mne_connectivity/viz/circle.py
```diff
@@ -151,10 +151,9 @@ def plot_connectivity_circle(
 
     if fig is not None or subplot is not None:
         warn(
-            "Passing a `fig` and `subplot` is deprecated and not be "
-            "supported after mne-connectivity version 0.4. Please "
-            "use the `ax` argument and pass a matplotlib axes object "
-            "with polar coordinates instead",
+            "Passing a `fig` and `subplot` is deprecated and not be supported after "
```
drammock, line 154:
```suggestion
            "Passing a `fig` and `subplot` is deprecated and not supported after "
```

drammock (approved):
just a few small improvements to suggest

mne_connectivity/tests/test_connectivity.py
tsbinns: Were a few other cases of this in the same file, so will change those too.

mne_connectivity/base.py
tsbinns: @drammock How's this?

https://github.com/mne-tools/mne-connectivity/blob/69c1116506eeb48d2c0aa7cf5d9473cbd227429f/mne_connectivity/base.py#L416-L420

larsoner:
Thanks @tsbinns

PR #208, opened 2024-06-20
[GSOC] Add decomposition plotting

tsbinns:
As described in #182.

This adds support for visualising spatial filters and patterns as topomaps with the `plot_filters` and `plot_patterns` method of the `CoherencyDecomposition` class. Very similar to the methods of the same name in the `CSP` class of MNE's decoding module.

Unit tests added as well as a short example on how these methods can be used and the figures interpreted.


mne_connectivity/decoding/decomposition.py
```diff
@@ -465,3 +469,306 @@ def get_transformed_indices(self):
...
+        _validate_type(info, Info, "`info`", "mne.Info")
+        if components is None:
+            components = np.arange(self.n_components)
+
+        # plot seeds and targets
+        figs = []
+        for group_idx, group_name in zip([0, 1], ["Seeds", "Targets"]):
+            # create info for seeds/targets
+            group_info = deepcopy(info)
+            group_info = pick_info(group_info, self.indices[group_idx], copy=False)
```
larsoner, line 731: `mne.Info` overrides the default `dict.copy` to do a deepcopy so you could do
```
            group_info = info.copy()
            group_info = pick_info(group_info, self.indices[group_idx], copy=False)
```
but even this can be made more compact just by using the default `copy=True` of `pick_info`
```suggestion
            group_info = pick_info(info, self.indices[group_idx])
```

mne_connectivity/decoding/decomposition.py
```diff
@@ -465,3 +469,306 @@ def get_transformed_indices(self):
...
+                    if name_format is None
+                    else name_format,
+                    nrows=nrows,
+                    ncols=ncols,
+                    show=False,  # set Seeds/Targets suptitle first
+                )
+            )
+            figs[-1].suptitle(group_name)  # differentiate seeds from targets
+            if show:
+                figs[-1].show()
```
larsoner, line 772: You should use `plt_show(fig=figs[-1])` from `mne.utils` as it does some stuff with warnings IIRC and also deals with the conditional
```suggestion
            plt_show(show, fig=figs[-1])
```

larsoner (commented):
Just some minor comments, the rendered example looks nice!

@wmvanvliet want to take a look?

examples/decoding/cohy_decomposition_plotting.py
```diff
@@ -0,0 +1,170 @@
...
+# connectivity**. Accordingly, keep in mind that the filters and patterns are not a
+# replacement for source reconstruction, as without this the patterns will still only
+# tell you about the spatial contributions of sensors, not underlying brain regions,
+# to connectivity.
+
+########################################################################################
+# Generating the filters and patterns
+# -----------------------------------
+# We will first load some example MEG data which we will generate the spatial filters
+# and patterns for, and divide it into epochs.
```
wmvanvliet, line 71: Explain briefly what kind of data this is? What paradigm?

mne_connectivity/utils/docs.py
```diff
@@ -250,6 +253,224 @@
...
+docdict["info_decoding_plotting"] = """
+info : mne.Info
+    Information about the sensors of the data which has been decomposed, such as that
+    coming from an :class:`mne.Epochs` object.
+"""
+
+# Topomaps
+docdict["components_topomap"] = """
+components : int | array of int | None (default None)
+    The components to plot. If `None`, all components are shown.
```
wmvanvliet, line 266:
```suggestion
    The components to plot. If ``None``, all components are shown.
```

wmvanvliet (changes requested):
Super small nitpicks. This looks great! The tutorial is informative and will be a nice addition to the docs.

mne_connectivity/decoding/decomposition.py
tsbinns: Done!

mne_connectivity/decoding/decomposition.py
tsbinns: Done!

examples/decoding/cohy_decomposition_plotting.py
tsbinns: I might be blind, but I didn't see the `fieldtrip_cmc` data described in the MNE docs with the other open datasets. From the name I'm guessing it has something to do with the data described here: https://www.fieldtriptoolbox.org/tutorial/coherence/

Updated to this:
https://github.com/mne-tools/mne-connectivity/blob/7709a329172c2e2e9b2f2048134ecacebe436ad3/examples/decoding/cohy_decomposition_plotting.py#L70-L72

tsbinns:
Thanks both for the feedback! That should be everything addressed.

adam2392:
Vey slick figures :)

examples/decoding/cohy_decomposition_plotting.py
adam2392: Perhaps it is worth adding a link to that specific tutorial to "see for more information"?

I always find it useful as a use to be able to conceptually tie w/e data we're looking at and the question to why we're using a specific computational method.

tsbinns: @adam2392 Happy to include a link to the ft example. Just FYI there are a few differences between theirs and ours:
- we look at connectivity between MEG sensors using icoh
- they look at connectivity between MEG and EMG using coh

But it still explains the data a bit more so can add this

tsbinns: Done!

larsoner (approved):
Thanks @tsbinns !

PR #210, opened 2024-06-28
Phase-slope index using spectral_connectivity_time instead of spectral_connectivity_epochs

seqasim:
Thanks for contributing. If this is your first time,
make sure to read [contributing.md](https://github.com/mne-tools/mne-connectivity/blob/main/CONTRIBUTING.md)

PR Description
--------------

Currently, the [function for computing phase-slope index](https://mne.tools/mne-connectivity/stable/generated/mne_connectivity.phase_slope_index.html) is hard-coded to be computed over epochs to produce time-resolved PSI, and there's no option for computing the phase-slope index over time to produce PSI per epoch. This would be useful if you think the average directionality between two signals over the entire timecourse is informative. You could compute this by binning your trials into categories (i.e. good trials v. bad trials) and then computing PSI, but there's no way to compute PSI per epoch and relate it to continuous trial-level predictors. 

Here, I add a new function `phase_slope_index_time` to `mne_connectivity.effective` that compute PSI over time. I should note that I wasn't able to test this function directly as I'm having trouble getting the forked repo installed, in editable mode, with all necessary dependencies, but have written a test function in `tests.test_effective` to facilitate this.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_connectivity/effective.py
+++ b/mne_connectivity/effective.py
@@ -257,11 +257,10 @@ def phase_slope_index_time(data,
                            mt_bandwidth=None,
                            freqs=None,
                            n_cycles=7, 
-                           average=True,
                            padding=0,
                            n_jobs=1,
                            ):
-    """Compute the Phase Slope Index (PSI) connectivity measure.
+    """Compute the Phase Slope Index (PSI) connectivity measure across time rather than epochs.
 
     The PSI is an effective connectivity measure, i.e., a measure which can
     give an indication of the direction of the information flow (causality).
@@ -307,27 +306,14 @@ def phase_slope_index_time(data,
     fmax : float | tuple of float
         The upper frequency of interest. Multiple bands are dedined using
         a tuple, e.g. (13., 30.) for two band with 13Hz and 30Hz upper freq.
-    tmin : float | None
-        Time to start connectivity estimation.
-    tmax : float | None
-        Time to end connectivity estimation.
     mt_bandwidth : float | None
         The bandwidth of the multitaper windowing function in Hz.
         Only used in 'multitaper' mode.
-    mt_adaptive : bool
-        Use adaptive weights to combine the tapered spectra into PSD.
-        Only used in 'multitaper' mode.
-    mt_low_bias : bool
-        Only use tapers with more than 90 percent spectral concentration within
-        bandwidth. Only used in 'multitaper' mode.
-    cwt_freqs : array
+    freqs : array
         Array of frequencies of interest. Only used in 'cwt_morlet' mode.
-    cwt_n_cycles : float | array of float
+    n_cycles : float | array of float
         Number of cycles. Fixed number or one per frequency. Only used in
         'cwt_morlet' mode.
-    block_size : int
-        How many connections to compute at once (higher numbers are faster
-        but require more memory).
     n_jobs : int
         How many epochs to process in parallel.
     %(verbose)s
@@ -335,8 +321,7 @@ def phase_slope_index_time(data,
     Returns
     -------
     conn : instance of Connectivity
-        Computed connectivity measure(s). Either a
-        ``SpectralConnnectivity``, or ``EpochSpectralConnectivity``
+        Computed connectivity measure(s). ``EpochSpectralConnectivity``
         container. The shape of each array is
         (n_signals ** 2, n_bands, n_epochs)
         when "indices" is None, or
@@ -351,7 +336,7 @@ def phase_slope_index_time(data,
     ----------
     .. footbibliography::
     """
-    logger.info("Estimating phase slope index (PSI)")
+    logger.info("Estimating phase slope index (PSI) across time")
 
     # estimate the coherency
 
--- a/mne_connectivity/effective.py
+++ b/mne_connectivity/effective.py
@@ -7,8 +7,8 @@ import copy
 import numpy as np
 from mne.utils import logger, verbose
 
-from .base import SpectralConnectivity, SpectroTemporalConnectivity
-from .spectral import spectral_connectivity_epochs
+from .base import SpectralConnectivity, SpectroTemporalConnectivity, EpochSpectralConnectivity
+from .spectral import spectral_connectivity_epochs, spectral_connectivity_time
 from .utils import fill_doc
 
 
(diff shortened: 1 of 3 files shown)


tsbinns:
Just need to make sure that the new `cohy` method is described in the docstring for `spectral_connectivity_time()`:

https://github.com/mne-tools/mne-connectivity/blob/ef0a4842d70ff784cbb3b913fc46033f25710ea2/mne_connectivity/spectral/time.py#L72-L75

<br>

And also an equation entry like for `coh`:

https://github.com/mne-tools/mne-connectivity/blob/ef0a4842d70ff784cbb3b913fc46033f25710ea2/mne_connectivity/spectral/time.py#L245-L256

tsbinns:
A more general comment for the new `phase_slope_index_time`: Even if we are not averaging over epochs in the call to `spectral_connectivity_time()`, do you think it makes sense to add an `average` option which would allow averaging over epochs ***after*** PSI has been computed?

mne_connectivity/effective.py
```diff
@@ -243,3 +247,191 @@ def phase_slope_index(
...
+        Computed connectivity measure(s). ``EpochSpectralConnectivity``
+        container. The shape of each array is
+        (n_signals ** 2, n_bands, n_epochs)
+        when "indices" is None, or
+        (n_con, n_bands, n_epochs) 
+        when "indices" is specified and "n_con = len(indices[0])".
+
+    See Also
+    --------
+    mne_connectivity.EpochSpectralConnectivity
```
tsbinns, line 337: I'm wondering if it's good to reiterate that `spectral_connectivity_time()` is a relevant function. Though I realise you based this on the existing function docstring which doesn't link to `spec_conn_epochs()`.

mne_connectivity/effective.py
```diff
@@ -243,3 +247,191 @@ def phase_slope_index(
...
+    n_cycles : float | array of float
+        Number of cycles. Fixed number or one per frequency. Only used in
+        'cwt_morlet' mode.
+    n_jobs : int
+        How many epochs to process in parallel.
+    %(verbose)s
+
+    Returns
+    -------
+    conn : instance of Connectivity
```
tsbinns, line 327: #214 made a lot of these typehints more explicit (i.e. `EpochSpectralConnectivity` instead).

```suggestion
    conn : instance of EpochSpectralConnectivity
```

mne_connectivity/effective.py
```diff
@@ -243,3 +247,191 @@ def phase_slope_index(
...
+                           mode="multitaper",
+                           fmin=None,
+                           fmax=np.inf,
+                           mt_bandwidth=None,
+                           freqs=None,
+                           n_cycles=7,
+                           padding=0,
+                           n_jobs=1,
+                           ):
+    """Compute the Phase Slope Index (PSI) connectivity measure across time rather than epochs.
```
tsbinns, line 267: Just make sure you add a verbose parameter. That should get the tests running.

mne_connectivity/effective.py
```diff
@@ -243,3 +247,191 @@ def phase_slope_index(
...
+
+    The PSI is computed from the coherency (see spectral_connectivity_epochs),
+    details can be found in :footcite:`NolteEtAl2008`.
+
+    This function computes PSI over time from epoched data.
+    The data may consist of a single epoch.
+
+    Parameters
+    ----------
+    data : array-like, shape=(n_epochs, n_signals, n_times)
```
tsbinns, line 288: This should also include `Epochs` (again, I realise you took this from the existing function which doesn't have this, so should change there too).

mne_connectivity/tests/test_effective.py
```diff
@@ -39,3 +39,40 @@ def test_psi():
 
     assert np.all(conn_cwt.get_data() > 0)
     assert conn_cwt.shape[-1] == n_times
+
+
+def test_psi_time():
+    """Test Phase Slope Index (PSI) estimation across time."""
```
tsbinns, line 45: From a quick glance the test looks good to me, let's see if it passes once the `verbose` param is added to the new function.

tsbinns (commented):
So just pretty minor comments. There are a couple formatting things (mainly line length) that I would have thought `pre-commit` would sort (if you've set that up). Would recommend doing so otherwise getting that CI check to pass will be very tedious...

seqasim:
> Hi @seqasim, thanks for the PR and sorry for not getting back to you sooner on this. I have a few comments/suggestions which I'll post below, but it's a really good start!
> 
> If this is something you're still interested in working on, your contributions are of course very welcome, however I realise it's been a while. If you're not able to invest time into this anymore and you don't want your hard work to go to waste, I can pick things up and we can make sure the changes are added.
> 
> Just let us know. Cheers!

Hey, 

Thanks for reviewing things! It's super illuminating for me to see the details I missed. I sadly won't have time to follow through on this for a few months, so if you feel you can wrap it up quickly, please don't wait for me! But if not, I'm happy to pick it up with all your helpful comments later in the year.

mne_connectivity/effective.py
drammock: Also you shouldn't need both verbose and fill_doc decorators. Verbose will trigger the same things as fill_doc

tsbinns:
> But if not, I'm happy to pick it up with all your helpful comments later in the year.

Yeah sounds good, there's no rush!

mne_connectivity/effective.py
seqasim: Yes, I think this is a good idea.

seqasim:
I think I addressed everything here

tsbinns:
Thanks for the changes @seqasim! I am quite busy at the moment, but I will try to do another round of reviews soon.
Looking at the CI logs, seems like the failures come from the `verbose` and `fill_doc` decorators not being completely redundant here, but from a quick look I'm not sure why.

seqasim:
Ok, I think I resolved all failures. There were some dimension issues with the way I was extracting the cohy results, which are now fixed.

mne_connectivity/effective.py
tsbinns: @drammock Turns out `fill_doc` was needed since this is MNE-Connectivity's `fill_doc`, with access to MNE-Connectivity's `docdict`, whereas calling `verbose` only used the main MNE package's `fill_doc`.

mne_connectivity/tests/test_effective.py
```diff
@@ -39,3 +39,40 @@ def test_psi():
...
+
+    # # the measure is symmetric (sign flip)
+    # assert_array_almost_equal(
+    #     conn_2.get_data()[0, 0], -conn.get_data(output="dense")[1, 0, 0]
+    # )
+
+    freqs = np.arange(5.0, 20, 0.5)
+    conn_cwt = phase_slope_index_time(
+        data, mode="cwt_morlet", sfreq=sfreq, freqs=freqs, indices=indices
+    )
```
tsbinns, line 75: Including some of the things commented here would be nice to better reflect the checks happening for the regular PSI function.

So as well as checking that seed -> target connectivity is > 0, should also check that target -> seed connectivity is < 0 and they are identical but just sign-flipped.

mne_connectivity/effective.py
drammock: Ahh right, I don't think of that

tsbinns (commented):
Have pushed some changes to formatting and also documentation updates to reflect the `spec_conn_time` more. Also added the new function to the API file and main init file.

One thing which I think is still outstanding is an `average` parameter that controls averaging over epochs, since this behaviour can be controlled with `spec_conn_time`.
It should be a pretty simple case of just updating the documentation and adding the extra behaviour of packaging the PSI data in a `SpectralConnectivity` container (just like for `spec_conn_time`).
The current implementation of computing PSI from coherency should already be flexible w.r.t. an epoch dimension or not, but this could also be added as a check in the unit test.

I also still think it makes sense to expose the `spec_conn_time` smoothing options (`sm_times`, `sm_freqs`, and `sm_kernel`).
This would just be a simple case of just updating the docstring.

Once that's added and the other point about the unit test being extended, I think this would be ready!

seqasim:
Hi @tsbinns , 

I finally got around to making these minor edits you requested! 

Salman

tsbinns:
Pushed a few commits:
- Updated the docs/params to better match `spec_conn_time`
- Updates the docs of the original and new time implementations to better match one another
- Refactored PSI computation to reduce repeated code between original and new time implementations
- Updated PSI time unit tests that the properties of results are as expected, including a check that frequency band-specific info is captured
- Move epoch averaging checks into a separate unit test
- Updated the existing PSI example to mention that estimates over time can also be computed with the new function

@drammock I'm happy with the state of this PR now. Please could I get your thoughts on whether we are good to merge.

@seqasim thanks for the extra changes, and sorry for the delay in reviewing!

PR #214, opened 2024-07-10
[DOC] Update functon-container cross-references and fix incorrect types

tsbinns:
Discussed partially in call with @larsoner & @wmvanvliet.

There is a `See Also` section of the docstrings where connectivity functions link to the connectivity containers they return, and the container classes like to the connectivity functions they are returned from. These references were not present everywhere, and I wanted to make them consistent.

However, I found there were existing references where functions were linking to classes they did not return, and classes were linking to functions they were not returned from. There were also cases where the types of container classes returned in connectivity function docstrings were wrong.

There is also an inconsistency where the `Connectivity` class is sometimes used as an indicator of the generic `BaseConnectivity` class, and sometimes used to refer to the `Connectivity` class specifically. Just thought this could be a potential point of confusion. E.g. `Connectivity` is listed as the sole accepted class type in `plot_connectivity_circle` when anything inheriting from `BaseConnectivity` is allowed.
In this case I didn't change anything because I wasn't sure if others consider it problematic.

### DIFF
--- a/mne_connectivity/base.py
+++ b/mne_connectivity/base.py
@@ -888,7 +888,9 @@ class SpectralConnectivity(BaseConnectivity, SpectralMixin):
 
     See Also
     --------
+    mne_connectivity.phase_slope_index
     mne_connectivity.spectral_connectivity_epochs
+    mne_connectivity.spectral_connectivity_time
     """
 
     expected_n_dim = 2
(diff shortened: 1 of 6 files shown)


tsbinns:
Were also some cases where `noqa` tags could be removed from docstrings thanks to the longer line lengths.

wmvanvliet (approved):

larsoner:
@wmvanvliet I was confused why your approval didn't satisfy the merge requirements then I realized you didn't have write permissions on this repo. Now you're an admin 🚀 

In it goes, thanks @tsbinns !

PR #217, opened 2024-07-25
[MAINT] Add `plot_sensors_connectivity` data shape check

tsbinns:
PR Description
--------------

Discussed in call with @larsoner.

`plot_sensors_connectivity` expects data to be a 2D (channels x channels) array, however the input shape is never checked. This PR adds such a check and corresponding unit tests.

No similar problem for `plot_connectivity_circle`, as data shape is checked here.

### DIFF
--- a/mne_connectivity/viz/_3d.py
+++ b/mne_connectivity/viz/_3d.py
@@ -57,6 +57,11 @@ def plot_sensors_connectivity(
 
     renderer = _get_renderer(size=(600, 600), bgcolor=(0.5, 0.5, 0.5))
 
+    if con.ndim != 2 or con.shape[0] != con.shape[1]:
+        raise ValueError(
+            "Connectivity data must be a 2D array of shape (channels, channels)"
+        )
+
     picks = _picks_to_idx(info, picks)
     if len(picks) != len(con):
         raise ValueError(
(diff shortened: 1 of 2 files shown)


mne_connectivity/viz/_3d.py
```diff
@@ -57,6 +57,11 @@ def plot_sensors_connectivity(
 
     renderer = _get_renderer(size=(600, 600), bgcolor=(0.5, 0.5, 0.5))
 
+    if con.ndim != 2 or con.shape[0] != con.shape[1]:
+        raise ValueError(
+            "Connectivity data must be a 2D array of shape (channels, channels)"
```
larsoner, line 62:
```suggestion
            "Connectivity data must be a 2D array of shape (n_channels, n_channels), "
            f"got shape {con.shape}"
```

mne_connectivity/viz/tests/test_3d.py
```diff
@@ -27,6 +27,10 @@ def test_plot_sensors_connectivity(renderer):
     info = raw.info
     with pytest.raises(TypeError, match="must be an instance of Info"):
         plot_sensors_connectivity(info="foo", con=con, picks=picks)
+    with pytest.raises(ValueError, match="Connectivity data must be a 2D array"):
+        plot_sensors_connectivity(info=info, con=np.expand_dims(con, 3), picks=picks)
+    with pytest.raises(ValueError, match="2D array of shape (channels, channels)"):
```
larsoner, line 32:
```suggestion
    with pytest.raises(ValueError, match="2D array of shape (n_channels, n_channels)"):
```

larsoner:
Just tweaked the error message a tiny bit, marking for merge-when-green, thanks in advance @tsbinns !

PR #218, opened 2024-07-25
[DOC] Switch Granger causality to expected failing example

tsbinns:
PR Description
--------------

Something that was discussed a while ago with @larsoner when the Granger causality example was first written but had to be postponed until new config options were added to sphinx-gallery:
- The GC example ends with a `try except` statement that is expected to fail
- It was suggested to remove the `try except` and just allow the error to be raised as long as the example is marked in sphinx as an expected failing example
- The problem was that expected failing examples automatically get a thumbnail with the word "BROKEN" which could not be overridden, and I felt this gave the wrong impression that there was a problem with the whole example
- In the new sphinx-gallery v0.17, support was added allowing this behaviour to be changed such that we can allow the GC example to fail while still giving it a proper thumbnail

Also a very small change to the decomposition docstring I overlooked before where a `None` wasn't being picked up as a type for intersphinx to link to.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -342,6 +342,7 @@ sphinx_gallery_conf = {
     "matplotlib_animations": True,
     "compress_images": ("images", "thumbnails"),
     "image_scrapers": scrapers,
+    "expected_failing_examples": ["../examples/granger_causality.py"],
 }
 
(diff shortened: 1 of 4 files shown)


larsoner:
Looks great, thanks @tsbinns !

PR #220, opened 2024-08-01
[GSOC] Add support for `EpochsSpectrum` objects in `spectral_connectivity_epochs`

tsbinns:
Precursor for adding surrogate connectivity stats testing.

Allows Fourier coefficients stored in `EpochsSpectrum` objects to be passed in as `data` to `spectral_connectivity_epochs()`. Support for storing Fourier coeffs. in these objects is currently a WIP: https://github.com/mne-tools/mne-python/pull/12747, so have had to modify the CircleCI tests to allow this.

### DIFF
--- a/examples/sensor_connectivity.py
+++ b/examples/sensor_connectivity.py
@@ -12,6 +12,8 @@ are used which produces strong connectvitiy in the right occipital sensors.
 #
 # License: BSD (3-clause)
 
+# %%
+
 import os.path as op
 
 import mne
@@ -22,7 +24,8 @@ from mne_connectivity.viz import plot_sensors_connectivity
 
 print(__doc__)
 
-###############################################################################
+# %%
+
 # Set parameters
 data_path = sample.data_path()
 raw_fname = op.join(data_path, "MEG", "sample", "sample_audvis_filt-0-40_raw.fif")
@@ -41,7 +44,7 @@ picks = mne.pick_types(
 )
 
 # Create epochs for the visual condition
-event_id, tmin, tmax = 3, -0.2, 1.5  # need a long enough epoch for 5 cycles
+event_id, tmin, tmax = 3, -0.2, 1.5  # want a long enough epoch for 5 cycles
 epochs = mne.Epochs(
     raw,
     events,
@@ -52,24 +55,18 @@ epochs = mne.Epochs(
     baseline=(None, 0),
     reject=dict(grad=4000e-13, eog=150e-6),
 )
+epochs.load_data().pick("grad")  # just keep MEG and no EOG now
 
-# Compute connectivity for band containing the evoked response.
-# We exclude the baseline period:
-fmin, fmax = 4.0, 9.0
-sfreq = raw.info["sfreq"]  # the sampling frequency
+# Compute Fourier coefficients for the epochs (returns an EpochsSpectrum object)
+# (storing Fourier coefficients in EpochsSpectrum objects requires MNE >= 1.8)
 tmin = 0.0  # exclude the baseline period
-epochs.load_data().pick_types(meg="grad")  # just keep MEG and no EOG now
+spectrum = epochs.compute_psd(method="multitaper", tmin=tmin, output="complex")
+
+# Compute connectivity for the frequency band containing the evoked response
+# (passing EpochsSpectrum objects as data requires MNE-Connectivity >= 0.8)
+fmin, fmax = 4.0, 9.0
 con = spectral_connectivity_epochs(
-    epochs,
-    method="pli",
-    mode="multitaper",
-    sfreq=sfreq,
-    fmin=fmin,
-    fmax=fmax,
-    faverage=True,
-    tmin=tmin,
-    mt_adaptive=False,
-    n_jobs=1,
+    data=spectrum, method="pli", fmin=fmin, fmax=fmax, faverage=True, n_jobs=1
 )
 
 # Now, visualize the connectivity in 3D:
--- a/mne_connectivity/spectral/epochs.py
+++ b/mne_connectivity/spectral/epochs.py
@@ -406,8 +406,8 @@ def _compute_spectra(
                 if mode == "multitaper":
                     weights = np.sqrt(eigvals)[np.newaxis, :, np.newaxis]
                 else:
-                    # hack to so we can sum over axis=-2
-                    weights = np.array([1.0])[:, None, None]
+                    # hack to so we can sum over axis=-2 (tapers dim)
+                    weights = np.ones((1, 1, 1))
 
                 if accumulate_psd:
                     _this_psd = _psd_from_mt(this_x_t, weights)
@@ -516,7 +516,7 @@ def _epoch_spectral_connectivity(
         x_t = np.array(data)[:, sig_idx, ..., freq_mask]
         if weights is None:  # also assumes no tapers dim
             x_t = np.expand_dims(x_t, axis=2)  # CSD construction expects a tapers dim
-            weights = np.array([1.0])[:, None, None]
+            weights = np.ones((1, 1, 1))
         if accumulate_psd:
             this_psd = _psd_from_mt(x_t, weights)
     else:  # compute spectral info from scratch
--- a/mne_connectivity/spectral/epochs.py
+++ b/mne_connectivity/spectral/epochs.py
@@ -19,6 +19,11 @@ from mne.time_frequency.multitaper import (
     _psd_from_mt,
     _psd_from_mt_adaptive,
 )
+from mne.time_frequency.spectrum import (
+    BaseSpectrum,
+    EpochsSpectrum,
+    EpochsSpectrumArray,
+)
 from mne.time_frequency.tfr import cwt, morlet
 from mne.utils import _arange_div, _check_option, _time_mask, logger, verbose, warn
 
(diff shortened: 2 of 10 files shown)


tsbinns:
@larsoner I updated an existing test like you suggested to show this new behaviour, with just a small description in the comments:

https://github.com/mne-tools/mne-connectivity/blob/6debece5f02f6d3f393c9191f4d98f36159c3c20/examples/sensor_connectivity.py#L60-L70

Also cleaned up an outdated use of `pick_types()`

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -467,6 +470,96 @@ def test_spectral_connectivity(method, mode):
     assert out_lens[0] == 10
 
 
+# Fourier coeffs in Spectrum objects added in MNE v1.8.0
+@pytest.mark.skipif(
+    Version(mne.__version__) <= Version("1.7.1"), reason="Requires MNE v1.8.0 or higher"
+)
```
tsbinns, line 476: Would ideally have checked version against 1.8, but dev versions are considered earlier than official releases. Will need to change this when MNE 1.8 is released.

larsoner: In `mne.utils.check_version` actually we treat dev versions differently, if you use that then it should do what you want

larsoner:
```
>>> import mne
>>> mne.__version__
'1.8.0.dev123+g790be2d0f'
>>> mne.utils.check_version("mne", "1.8")
True
```

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -467,6 +470,96 @@ def test_spectral_connectivity(method, mode):
...
+    assert_array_less(con.get_data()[:, freqs_noise], 0.2)
+
+
+# TODO: Add general test for error catching for spec_conn_epochs
+# Fourier coeffs in Spectrum objects added in MNE v1.8.0
+@pytest.mark.skipif(
+    Version(mne.__version__) <= Version("1.7.1"), reason="Requires MNE v1.8.0 or higher"
+)
+def test_spectral_connectivity_epochs_spectrum_input_error_catch():
+    """Test spec_conn_epochs catches error with EpochsSpectrum data as input."""
```
tsbinns, line 543: Realised there isn't a standalone test for catching errors from spec_conn_epochs, so have added one for checking spectrum inputs.

Have included a `TODO` tag that maybe the error catching should be disentangled from `test_spectral_connectivity` in future.

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -527,7 +620,7 @@ def test_spectral_connectivity_epochs_multivariate(method, n_components):
     freqs = np.array(con.freqs)
     freqs_con = (freqs >= fstart) & (freqs <= fend)
     freqs_noise = (freqs < fstart - trans_bandwidth * 2) | (
-        freqs > fend + -trans_bandwidth * 2
+        freqs > fend + trans_bandwidth * 2
     )
```
tsbinns, line 624: Little bug fix in testing...

mne_connectivity/spectral/time.py
```diff
@@ -165,6 +165,8 @@ def spectral_connectivity_time(
         ``n_components`` must be <= 3. If `None`, the number of components equal to the
         minimum rank of the seeds and targets is extracted (see the ``rank`` parameter).
         Only used if ``method`` contains any of ``['cacoh', 'mic']``.
+
+        .. versionadded:: 0.8
```
tsbinns, line 169: Forgot to add `versionadded` tag for spec_conn_time `components` param

mne_connectivity/spectral/tests/test_spectral.py
tsbinns: Have now added a proper reminder in code to do this

larsoner: No I mean you can do it now, in this PR. If you do
```
pytest.mark.skipif(not check_version("mne", "1.8"), ...)
```
it will work now and once 1.8 is released just fine, no need for any "change this later" comment

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -467,6 +470,98 @@ def test_spectral_connectivity(method, mode):
...
+    assert_array_less(0.6, con.get_data()[:, freqs_con])
+    assert_array_less(con.get_data()[:, freqs_noise], 0.2)
+
+
+# TODO: Add general test for error catching for spec_conn_epochs
+# Fourier coeffs in Spectrum objects added in MNE v1.8.0
+# FIXME: Update to reference MNE 1.8 instead of 1.7.1 when 1.8 released
+@pytest.mark.skipif(
+    Version(mne.__version__) <= Version("1.7.1"), reason="Requires MNE v1.8.0 or higher"
+)
```
larsoner, line 543:
```suggestion
# TODO: Add general test for error catching for spec_conn_epochs
@pytest.mark.skipif(not check_version("mne", "1.8"), reason="Requires MNE v1.8.0 or higher"
)
```

tsbinns:
Hmm, seems like calling `pip install git+https://github.com/tsbinns/mne-python@complex_spectrum` ends up installing a dev branch of MNE 1.4

mne_connectivity/spectral/tests/test_spectral.py
tsbinns: Sorry, I didn't refresh the page before adding that reminder comment, will switch to your solution.

larsoner:
> ends up installing a dev branch of MNE 1.4

Hah! This is a little tricky, you need to locally on your machine 
```
$ git fetch upstream --tags
$ git push origin --tags
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To github.com:/larsoner/mne-python.git
 * [new tag]               v1.6.1 -> v1.6.1
 * [new tag]               v1.7.0 -> v1.7.0
 * [new tag]               v1.7.1 -> v1.7.1
```
because it needs all the recent tags *in your fork* to get the version right when when installing with pip + git.

tsbinns:
That looks better:
image

Never would've figure that out. Thanks!

tsbinns:
I fixed something for the Welch/Fourier mode I overlooked in my previous implementation when I was refactoring the code. Turns out that changed some of the connectivity values so the thresholds in my test for checking that connectivity was present in simulated data stopped working. Will push something to fix that.

I'm confident my implementation is working properly, because I can get the difference between the connectivity results from the Welch `EpochsSpectrum` and Fourier mode of `spec_conn_epochs` down to at most 0.006.

mne_connectivity/spectral/tests/test_spectral.py
tsbinns: Have switched to that now

tsbinns:
Sorry, not my finest PR but tests are sorted now

mne_connectivity/spectral/epochs.py
```diff
@@ -387,60 +510,30 @@ def _epoch_spectral_connectivity(
...
-            _this_psd = (this_x_t * this_x_t.conj()).real
-
-        x_t.append(this_x_t)
+    if spectrum_computed:  # use existing spectral info
+        # XXX: Will need to distinguish time-resolved spectra here if support added
+        # Select signals & freqs of interest (flexible indexing for optional tapers dim)
+        x_t = np.array(data)[:, sig_idx, ..., freq_mask]
+        if weights is None:  # also assumes no tapers dim
+            x_t = np.expand_dims(x_t, axis=2)  # CSD construction expects a tapers dim
+            weights = np.array([1.0])[:, None, None]
```
larsoner, line 519: Equivalent I think but a little more explicit about resulting dimensionality
```suggestion
            weights = np.ones((1, 1, 1))
```

mne_connectivity/spectral/epochs.py
```diff
@@ -634,14 +727,28 @@ def spectral_connectivity_epochs(
...
+            data = [(arr_0, stc_0), (arr_1, stc_1), (arr_2, stc_2)]
+
+        which corresponds to 3 epochs where ``arr_*`` is an array with the same number
+        of time points as ``stc_*``. Data can also be a `list`/:term:`generator` of
+        arrays, ``shape (n_signals, n_times)``, or a `list`/:term:`generator` of
+        `~mne.SourceEstimate` or `~mne.VolSourceEstimate` objects.
+
+        .. versionchanged:: 0.8
+           Fourier coefficients stored in an `~mne.time_frequency.EpochsSpectrum` or
+           `~mne.time_frequency.EpochsSpectrumArray` object can also be passed in as
```
larsoner, line 750: At some point I think we decided to go back to being explicit about these refs since its behavior depends on `default_role` (and can silently fail to link). So here and elsewhere it would be better to do:
```suggestion
           Fourier coefficients stored in an :class:`~mne.time_frequency.EpochsSpectrum` or
           :class:`~mne.time_frequency.EpochsSpectrumArray` object can also be passed in as
```

larsoner (commented):
Just some minor stuff, otherwise looks reasonable!

mne_connectivity/spectral/epochs.py
tsbinns: Okay, will add these.
Think I've also been lazy with the decomposition stuff as well, so will open an issue as a reminder to myself to fix.

mne_connectivity/spectral/epochs.py
tsbinns: Have changed there and also in `_compute_spectra()`

mne_connectivity/spectral/epochs.py
tsbinns: spec_conn_epochs should all be explicit now

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -31,8 +33,14 @@ def _check_rank_input(rank, data, indices):
             if "copy" in inspect.getfullargspec(data.get_data).kwonlyargs:
                 kwargs["copy"] = False
             data_arr = data.get_data(**kwargs)
-        else:
-            data_arr = data
+        elif isinstance(data, (EpochsSpectrum, EpochsSpectrumArray)):
+            # Spectrum objs will drop bad channels, so specify picking all channels
+            data_arr = data.get_data(picks=np.arange(data.info["nchan"]))
```
tsbinns, line 38: Realised that `Spectrum.get_data(picks=None)` returns only good channels, but the same for `Epochs` returns all channels, so need to explicitly pick all channels from `Spectrum` objects to keep behaviour consistent.

mne_connectivity/spectral/epochs_multivariate.py
```diff
@@ -31,8 +33,14 @@ def _check_rank_input(rank, data, indices):
...
+            # Convert to power (and aggregate over tapers) before computing rank
+            if "taper" in data._dims:
+                data_arr = _psd_from_mt(data_arr, data.weights)
+            else:
+                data_arr = (data_arr * data_arr.conj()).real
```
tsbinns, line 43: Also forgot to aggr. over tapers before checking the rank (otherwise we have a tapers dim in the singular values, and I'm not sure how to handle that).

Converting to power isn't explicitly necessary I guess, but I at least know it's a proper way to handle the tapers. And if we're doing that for multitaper coeffs, should keep it consistent and convert to power for Fourier/Welch as well. Also they're just simple matmuls, so not a huge overhead in this additional step.

tsbinns:
Will update spectrum tests next week with multivar. methods to make sure this gets coverage

tsbinns:
Have expanded test coverage, should make things more robust.

tsbinns:
Switching CI back to upstream since https://github.com/mne-tools/mne-python/pull/12747 was merged.

examples/sensor_connectivity.py
```diff
@@ -52,24 +55,18 @@
...
-    mode="multitaper",
-    sfreq=sfreq,
-    fmin=fmin,
-    fmax=fmax,
-    faverage=True,
-    tmin=tmin,
-    mt_adaptive=False,
-    n_jobs=1,
+    data=spectrum, method="pli", fmin=fmin, fmax=fmax, faverage=True, n_jobs=1
 )
```
larsoner, line 70: I checked by eye this example and it did indeed look the same which is good!

https://output.circle-artifacts.com/output/job/af5d79d3-61cd-4dd6-af69-a46142b3ce5d/artifacts/0/dev/auto_examples/sensor_connectivity.html#sphx-glr-auto-examples-sensor-connectivity-py

larsoner (approved):
Appears to be all good so in it goes, thanks @tsbinns !

examples/sensor_connectivity.py
tsbinns: Yeah, everything I tried with multitaper was identical, just some minor deviations for the Fourier/Welch.

Issue #148, opened 2023-09-22
plot_sensors_connectivity fails when strongest conns are spatially close


drammock:
#### Describe the bug
plot_sensors_connectivity() fails with a cryptic error message when all of the `n_con` strongest connections occur between sensor pairs that are less than 5cm apart.

#### Steps to reproduce
in this example:

https://mne.tools/mne-connectivity/dev/auto_examples/sensor_connectivity.html

and in the call to `spectral_connectivity_epochs` change method from `pli` to `coh`. Running the modified example will reproduce the error.  Passing in `n_con=50` will allow the plot to generate successfully.

#### Expected results
a plot of the most strongly connected sensors

#### Actual results
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File /opt/mne/connectivity/examples/sensor_connectivity.py:59
     54 con = spectral_connectivity_epochs(
     55     epochs, method='coh', mode='multitaper', sfreq=sfreq, fmin=fmin, fmax=fmax,
     56     faverage=True, tmin=tmin, mt_adaptive=False, n_jobs=1)
     58 # Now, visualize the connectivity in 3D:
---> 59 plot_sensors_connectivity(
     60     epochs.info,
     61     con.get_data(output='dense')[:, :, 0])

File /opt/mne/connectivity/mne_connectivity/viz/_3d.py:90, in plot_sensors_connectivity(info, con, picks, cbar_label, n_con, cmap)
     87 con_val = np.array(con_val)
     89 # Show the connections as tubes between sensors
---> 90 vmax = np.max(con_val)
     91 vmin = np.min(con_val)
     92 for val, nodes in zip(con_val, con_nodes):

File <__array_function__ internals>:180, in amax(*args, **kwargs)

File /opt/mambaforge/envs/mnedev/lib/python3.10/site-packages/numpy/core/fromnumeric.py:2793, in amax(a, axis, out, keepdims, initial, where)
   2677 @array_function_dispatch(_amax_dispatcher)
   2678 def amax(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
   2679          where=np._NoValue):
   2680     """
   2681     Return the maximum of an array or maximum along an axis.
   2682 
   (...)
   2791     5
   2792     """
-> 2793     return _wrapreduction(a, np.maximum, 'max', axis, None, out,
   2794                           keepdims=keepdims, initial=initial, where=where)

[...]

PR #221, opened 2024-08-01
[BUG, ENH] Improve error checking of `plot_sensors_connectivity` and add a `min_distance` parameter

tsbinns:
Addresses #148, discussed in call with @larsoner.

Have gone with options (2) & (3) suggested above. This involves:
- An option to specify the minimum distance required between sensors for a connection to be plotted (default was 5 cm, which has been retained).
- A check that if no connections survive the minimum distance requirement, a more descriptive error message is raised.
    - The message suggests increasing the number of connections being plotted or decreasing the minimum distance (would help in cases like https://mne.discourse.group/t/problem-with-different-methods-in-sensor-connectivity-example/7548/4)
    - The message also suggests checking that the coordinates of channels are not NaNs (would help in cases like https://mne.discourse.group/t/error-while-trying-to-call-plot-sensors-connectivity/9128)

### DIFF
--- a/mne_connectivity/viz/_3d.py
+++ b/mne_connectivity/viz/_3d.py
@@ -20,7 +20,13 @@ FIDUCIAL_ORDER = (FIFF.FIFFV_POINT_LPA, FIFF.FIFFV_POINT_NASION, FIFF.FIFFV_POIN
 
 @fill_doc
 def plot_sensors_connectivity(
-    info, con, picks=None, cbar_label="Connectivity", n_con=20, cmap="RdBu"
+    info,
+    con,
+    picks=None,
+    cbar_label="Connectivity",
+    n_con=20,
+    cmap="RdBu",
+    min_distance=0.05,
 ):
     """Visualize the sensor connectivity in 3D.
 
@@ -37,9 +43,12 @@ def plot_sensors_connectivity(
     n_con : int
         Number of strongest connections shown. By default 20.
     cmap : str | instance of matplotlib.colors.Colormap
-        Colormap for coloring connections by strength. If :class:`str`, must
-        be a valid Matplotlib colormap (i.e. a valid key of
-        ``matplotlib.colormaps``). Default is ``"RdBu"``.
+        Colormap for coloring connections by strength. If :class:`str`, must be a valid
+        Matplotlib colormap (i.e. a valid key of `matplotlib.colormaps`). Default is
+        ``"RdBu"``.
+    min_distance : float
+        The minimum distance required between two sensors to plot a connection between
+        them, in meters. Default is 0.05 (i.e. 5 cm).
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


mne_connectivity/viz/_3d.py
```diff
@@ -37,9 +43,12 @@ def plot_sensors_connectivity(
...
     cmap : str | instance of matplotlib.colors.Colormap
-        Colormap for coloring connections by strength. If :class:`str`, must
-        be a valid Matplotlib colormap (i.e. a valid key of
-        ``matplotlib.colormaps``). Default is ``"RdBu"``.
+        Colormap for coloring connections by strength. If :class:`str`, must be a valid
+        Matplotlib colormap (i.e. a valid key of `matplotlib.colormaps`). Default is
+        ``"RdBu"``.
+    min_distance : float
+        The minimum distance required between two sensors to plot a connection between
+        them, in meters. Default is 0.05 (i.e. 5 cm).
```
tsbinns, line 51: Is sticking with the default unit of metres okay, or would it be more intuitive to use something smaller in this case, e.g. cm?

mne_connectivity/viz/_3d.py
```diff
@@ -20,7 +20,13 @@
 
 @fill_doc
 def plot_sensors_connectivity(
-    info, con, picks=None, cbar_label="Connectivity", n_con=20, cmap="RdBu"
+    info,
+    con,
+    picks=None,
+    cbar_label="Connectivity",
```
larsoner, line 26: We try to add `*` nowadays, maybe this is a good spot?
```suggestion
    *,
    cbar_label="Connectivity",
```

mne_connectivity/viz/_3d.py
larsoner, line 51: versionadded

larsoner (approved):
Looks good other than a couple tiny things

mne_connectivity/viz/_3d.py
tsbinns: I even used the American spelling to be nice

tsbinns: Done!

mne_connectivity/viz/_3d.py
```diff
@@ -37,9 +44,13 @@ def plot_sensors_connectivity(
...
-        Colormap for coloring connections by strength. If :class:`str`, must
-        be a valid Matplotlib colormap (i.e. a valid key of
-        ``matplotlib.colormaps``). Default is ``"RdBu"``.
+        Colormap for coloring connections by strength. If :class:`str`, must be a valid
+        Matplotlib colormap (i.e. a valid key of `matplotlib.colormaps`). Default is
+        ``"RdBu"``.
+    min_distance : float
+        The minimum distance required between two sensors to plot a connection between
+        them, in meters. Default is 0.05 (i.e. 5 cm).
+        .. versionadded:: 0.8
```
larsoner, line 53: Pretty sure for this to render properly it needs a newline in between
```suggestion

        .. versionadded:: 0.8
```

tsbinns: You're right, sorry, was too hasty

PR #223, opened 2024-08-05
[GSOC] Add surrogate data generation for significant connectivity estimation

tsbinns:
Follows from the changes in #220, so the diff will be quite smaller when that's merged.

In brief, adds a new function `make_surrogate_connectivity` in the `datasets` module (felt this was an appropriate place).

Since `datasets` is more populous now I thought it could make sense to clean things up and bring the tests from `make_signals_in_freq_bands` from the generic tests folder to one within the datasets folder (where the `make_surrogate_connectivity` tests also are).

Also added an example demonstrating the use of the new function.


examples/surrogate_connectivity.py
```diff
@@ -0,0 +1,351 @@
...
+# When generating surrogate data, it is important to distinguish non-evoked data (e.g.,
+# resting-state, pre/inter-trial data) from evoked data (where a stimulus is presented
+# or an action performed at a set time during each epoch). Critically, evoked data
+# contains a temporal structure that is consistent across epochs, and thus shuffling
+# epochs across channels will fail to adequately disrupt the covariance structure.
+#
+# Any connectivity estimates will therefore overestimate the baseline connectivity in
+# your data, increasing the likelihood of type II errors (see the Notes section of
+# :func:`~mne_connectivity.make_surrogate_data` for more information, and see the final
+# section of this example for a demonstration).
```
tsbinns, line 212: Anyone know how to link to a section header in the same file? I tried a couple things but couldn't figure it out.

mne_connectivity/datasets/surrogate.py
```diff
@@ -0,0 +1,141 @@
...
+    data (e.g., rest data, inter-trial data) and compare this to your true connectivity
+    estimates from the evoked data.
+
+    Regardless of whether you are working with evoked or non-evoked data, **you should
+    always compare true and surrogate connectivity estimates from epochs of the same
+    duration**. This will ensure that spectral information is captured with the same
+    accuracy in both sets of connectivity estimates. Ideally, **you should also compare
+    true and surrogate connectivity estimates from the same number of epochs** to avoid
+    biases from noise (fewer epochs gives noisier estimates) or finite sample sizes
+    (e.g., in coherency, phase-locking value, etc... :footcite:`VinckEtAl2010`).
```
tsbinns, line 70: Very long notes section, but like we discussed, there are important considerations as to what data surrogates should be generated from.

tsbinns:
Hmm, building docs is failing here on the new example (fine for me locally), but doesn't give a reason beyond

```
make: *** [Makefile:56: html] Killed
Exited with code exit status 2
```

Could it be that it doesn't like multiprocessing? I'm using `n_jobs=-1` to speed up connectivity computation on the multiple shuffles.

larsoner:
Often that's a sign that you've used too much memory, and decreasing the number of jobs can help

tsbinns:
Okay, even using 1/4 the CPU count ends up using too much memory (find it surprising with 36 cores and 70 GB). Just reverting to a single job and taking the hit on runtime.

tsbinns:
Switching CI back to upstream since https://github.com/mne-tools/mne-python/pull/12747 was merged.

examples/surrogate_connectivity.py
```diff
@@ -0,0 +1,351 @@
...
+# Generate surrogate data
+n_shuffles = 100  # recommended is >= 1,000; limited here to reduce compute time
+pretrial_surrogates = make_surrogate_data(
+    pretrial_coeffs, n_shuffles=n_shuffles, rng_seed=44
+)
+
+# Compute connectivity for surrogate data
+surrogate_con = []
+for shuffle_i, surrogate in enumerate(pretrial_surrogates):
+    print(f"Computing connectivity for shuffle {shuffle_i+1} of {n_shuffles}")
```
wmvanvliet, line 122:
```suggestion
for shuffle_i, surrogate in enumerate(pretrial_surrogates, 1):
    print(f"Computing connectivity for shuffle {shuffle_i} of {n_shuffles}")
```

examples/surrogate_connectivity.py
wmvanvliet: Perhaps something like this?

```
`Titles are targets, too`_.
```

larsoner: We do this in MNE for example with the `build-docs` and later `.. _build-docs:`

https://github.com/mne-tools/mne-python/blame/6da505591cd76bc7f6579154475c4abb2a5cbb95/doc/development/contributing.rst#L436

tsbinns: Ah okay, I was on the right track, just had some underscores in the wrong places. Thanks both!

examples/surrogate_connectivity.py
tsbinns: Changed here and later on in the example.

larsoner:
@wmvanvliet want to look and merge if you're happy?

adam2392 (commented):
This looks so great! thanks for all the hard work @tsbinns 

https://output.circle-artifacts.com/output/job/bafe351e-3009-46ea-92ce-863de6768196/artifacts/0/dev/auto_examples/surrogate_connectivity.html#sphx-glr-auto-examples-surrogate-connectivity-py

Few comments:

1. would it be great to actually print the pvalue too potentially besides just writing that it's <0.05?
2. When comparing significance across the frequency band, the major difference appears in the alpha band, but you test the beta band. Perhaps it's worth mentioning that you test the beta band specifically since the alpha band is clearly significantly different (and due to simplicity for running the example), but in practice one can test both. You do allude to this point in multiple comparisons discussion, but that's one question I had when reading through that section

tsbinns:
Comments should be addressed now @adam2392.

adam2392 (approved):
LGTM

tsbinns:
Hi @adam2392 & @larsoner, is there something you think is still missing from this I should be working on? I have some time this month if so.

PR #227, opened 2024-08-13
unpin pytest

drammock:
cf. https://github.com/mne-tools/mne-python/issues/12789

Locally for me all tests pass except for 3 symmetric_orth runs (2-False, 3-False, and 3-True). **They pass on pytest 7.4.4** so this needs digging into


larsoner:
Yes most `pytest>=8` failures I ran into in moving other repos over was the need to have multiple `pytest.warns` context managers

larsoner (approved):

PR #228, opened 2024-08-20
[MAINT] Upgrade from Python 3.9 to 3.10

tsbinns:
Unit tests using Python 3.9 failing since it's no longer supported by MNE, so have just switched these to Python 3.10.


tsbinns:
Also makes it UP038 compliant, as in https://github.com/mne-tools/mne-python/pull/12799

larsoner:
Thanks @tsbinns !

Issue #225, opened 2024-08-08
GSOC Cleanup: Add `EpochsTFR` support to `spectral_connectivity_epochs()` and `spectral_connectivity_time()`


tsbinns:
Discussed in call with @wmvanvliet & @larsoner. Most likely I would not be able to address this until ~November.

Part of the GSoC project was to add tools for generating surrogate connectivity data (#223). As this acts by shuffling Fourier coefficients, part of the work involved adding `EpochsSpectrum` objects as a valid data type for `spectral_connectivity_epochs()` (#220). However, this only addresses the Fourier/Welch and multitaper spectral modes of `spectral_connectivity_epochs()`, ignoring the Morlet wavelet approach, as well as `spectral_connectivity_time()` entirely.

To finalise the GSoC work, support should be added for `EpochsTFR` objects in `spectral_connectivity_epochs()` and `spectral_connectivity_time()`. Thankfully this should be easier than adding `EpochsSpectrum` was ([mne-python#12747](https://github.com/mne-tools/mne-python/pull/12747)), since `EpochsTFR` already support Fourier coefficients.

This would also go some way towards addressing #126 (only 1.5 years later...), since you could pass an `EpochsTFR` object with time-freq. Fourier coeffs. from the multitaper mode into `spectral_connectivity_epochs()`.

tsbinns:
Sorted in #232!

Issue #126, opened 2023-02-10
Add support for additional TFR methods


tsbinns:
Currently, `cwt_morlet` is the only TFR method supported, and I see there is a note in `epochs.py`: `# reminder to add alternative TFR methods`.

Would it be possible to add support for e.g. the [multitaper TFR method available in the `mne-base` package](https://mne.tools/dev/generated/mne.time_frequency.tfr_array_multitaper.html)?

Best,
Thomas!

adam2392:
Yes that is a good idea. A PR is welcome! :) Lmk if you need guidance

tsbinns:
Gonna close this as completed now, since with #232 `EpochsTFR` objects are supported by the spectral connectivity functions, allowing the multitaper TFR method to be used.

PR #232, opened 2024-09-06
[GSOC] Add `EpochsTFR` support to spectral connectivity functions

tsbinns:
WIP follow up of #225 to finalise GSoC work.

So far adds support for coefficients from `Epochs.compute_tfr(method="morlet", output="complex")` to `spectral_connectivity_epochs()`, equivalent to the `cwt_morlet` mode. Still to be done is adding support for `EpochsTFR` objects in `spectral_connectivity_time()`.

In addition to the Morlet approach, `spectral_connectivity_time()` supports the time-freq. multitaper mode. #126 could also be addressed by adding support for this to `spectral_connectivity_epochs()`, in the form of `EpochsTFR` objects.

However, when trying this I discovered a bug that prevents the time-freq. multitaper mode being computed from epoched data (https://github.com/mne-tools/mne-python/issues/12831), but that seems like an easy fix.

Will continue to work on this next week.


tsbinns:
Have now added support for `EpochsTFR` to `spectral_connectivity_time` (using a custom MNE branch with the bug in #12831 fixed).

mne_connectivity/spectral/time.py
```diff
@@ -784,51 +843,62 @@ def _spectral_connectivity(
...
-            half_nbw = mt_bandwidth / 2.0
-            n_tapers = int(np.floor(mt_bandwidth - 1))
             _, eigvals = dpss_windows(window_length, half_nbw, n_tapers, sym=False)
             weights[:, i, :] = np.sqrt(eigvals[:, np.newaxis])
             # weights have shape (n_tapers, n_freqs, n_times)
     else:
-        raise ValueError("Mode must be 'cwt_morlet' or 'multitaper'.")
-
-    out = np.squeeze(out, axis=0)
+        weights = None
```
tsbinns, line 899: Just some refactoring here to reduce code repetition with the new `EpochsTFR` support.

mne_connectivity/spectral/time.py
```diff
@@ -392,12 +420,40 @@ def spectral_connectivity_time(
...
+                if mt_bandwidth is None:
+                    warn(
+                        "`mt_bandwidth` is not specified; assuming 4.0 Hz was used to "
+                        "compute the TFR"
+                    )
+                if n_cycles == 7.0:
+                    warn(
+                        "`n_cycles` is the default value; assuming 7.0 was used to "
+                        "compute the TFR"
+                    )
```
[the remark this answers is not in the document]
larsoner: Maybe we should add the `mt_bandwidth` and `n_cycles` attributes to the `*TFR` objects?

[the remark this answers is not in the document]
tsbinns: It would be nice to have. I'm happy to open a PR if people think it's justified.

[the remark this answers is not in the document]
larsoner: Provenance is usually nice to track and if it's easy enough to add to the class then we might as well. Might be worth a quick MNE issue as we might want them as attributes like `tfr.n_cycles` but maybe better would be something like a dict of params like `tfr.estimation_parameters = {'mt_bandwidth': 4, 'n_cycles': 5}` so that the `tfr` namespace itself doesn't get too busy.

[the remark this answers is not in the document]
tsbinns: Sure, will open an issue first then.

tsbinns:
Resetting tests to main mne branch now that https://github.com/mne-tools/mne-python/pull/12842 is merged

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -546,25 +553,18 @@ def test_spectral_connectivity_epochs_spectrum_input(method, mode):
...
 
 
 # TODO: Add general test for error catching for spec_conn_epochs
 @pytest.mark.skipif(
     not check_version("mne", "1.8"), reason="Requires MNE v1.8.0 or higher"
 )
-def test_spectral_connectivity_epochs_spectrum_input_error_catch():
-    """Test spec_conn_epochs catches error with EpochsSpectrum data as input."""
+def test_spectral_connectivity_epochs_spectrum_tfr_input_error_catch():
+    """Test spec_conn_epochs catches errors with EpochsSpectrum/TFR data as input."""
```
drammock, line 567: why change the test name if not changing the function to actually test TFR input?

mne_connectivity/spectral/time.py
```diff
@@ -66,12 +72,23 @@ def spectral_connectivity_time(
...
-    data : array_like, shape (n_epochs, n_signals, n_times) | Epochs
-        The data from which to compute connectivity.
+    data : array_like, shape (n_epochs, n_signals, n_times) | ~mne.Epochs | ~mne.time_frequency.EpochsTFR
+        The data from which to compute connectivity. Can be epoched timeseries data as
+        an :term:`array-like` or :class:`~mne.Epochs` object, or Fourier coefficients
+        for each epoch as an :class:`~mne.time_frequency.EpochsTFR` object. If
+        timeseries data, the spectral information will be computed according to the
+        spectral estimation mode (see the ``mode`` parameter). If an
+        :class:`~mne.time_frequency.EpochsTFR` object, this spectral information will be
+        used and the ``mode`` parameter will be ignored.
```
drammock, line 82:
```suggestion
        :class:`~mne.time_frequency.EpochsTFR` object, existing spectral information
        will be used and the ``mode`` parameter will be ignored.
```

mne_connectivity/spectral/time.py
```diff
@@ -133,21 +150,26 @@ def spectral_connectivity_time(
...
-        frequency. The number of cycles ``n_cycles`` and the frequencies of
-        interest ``cwt_freqs`` define the temporal window length. For details,
-        see :func:`mne.time_frequency.tfr_array_morlet` documentation.
+        Number of cycles in the wavelet, either a fixed number or one per frequency. The
+        number of cycles ``n_cycles`` and the frequencies of interest ``freqs`` define
+        the temporal window length. For details, see
+        :func:`mne.time_frequency.tfr_array_multitaper` and
+        :func:`mne.time_frequency.tfr_array_morlet` documentation. If ``data`` is an
+        :class:`~mne.time_frequency.EpochsTFR` object computed with the ``multitaper``
+        method, this should match the value used to compute the TFR.
```
drammock, line 172: statements like this worry me, because they mean that it's easy for people to make mistakes.

For things that are ignored if you pass an `EpochsTFR` object, then (assuming the default param value is `None` or some other null/auto value) we at least have the option of checking whether non-default input to this function *matches* what we extract/infer from the object, and raising warnings/errors if they don't match. But for params that have int/float default values, or for quantities/settings that were used to create the TFR object but aren't stored with it, we can't even do that sanity check.

I don't want to throw a big monkey wrench into this PR, but an API like this makes me think that there should be separate functions for operating on MNE Objects and on arrays, and the shared code path should be a helper func rather than the user-facing function(s). Another possibility is 2 public-facing funcs and no shared helper; one of the public funcs would then call the other after suitable transformation of its inputs.  This could mean internally extracting arrays from objects and always operating on arrays, or it could mean internally promoting arrays to objects and always operating on objects.

Without looking closely I'm not sure how much work that would be / how (in)consistent it is with the rest of the mne_connectivity API.  I also apologize if this was discussed in advance and I missed/forgot about that discussion...  I've been rather underwater for the last several months with other responsibilities.  But I'd like to at least check with you and @larsoner (and maybe @adam2392 if he has time) before this becomes something that needs a deprecation cycle to change.

(side note re: the discussion elsewhere in this PR between @tsbinns and @larsoner about storing estimation params like `n_cycles` as part of the TFR object --- that helps prevent some mistakes, but I still have the feeling like this function is trying to do too much.)

mne_connectivity/spectral/time.py
```diff
@@ -369,12 +391,18 @@ def spectral_connectivity_time(
...
-    """
+    """  # noqa: E501
     events = None
     event_id = None
     # extract data from Epochs object
-    _validate_type(data, (np.ndarray, BaseEpochs), "`data`", "Epochs or a NumPy array")
-    if isinstance(data, BaseEpochs):
+    _validate_type(
+        data,
+        (np.ndarray, BaseEpochs, EpochsTFR, EpochsTFRArray),
```
drammock, line 400:
```suggestion
        (np.ndarray, BaseEpochs, EpochsTFR),
```

mne_connectivity/spectral/time.py
```diff
@@ -369,12 +391,18 @@ def spectral_connectivity_time(
...
+        "`data`",
+        "Epochs, EpochsTFR, or a NumPy array",
+    )
+    spectrum_computed = False
+    if isinstance(data, BaseEpochs | EpochsTFR | EpochsTFRArray):
```
drammock, line 405:
```suggestion
    if isinstance(data, BaseEpochs | EpochsTFR):
```

mne_connectivity/spectral/time.py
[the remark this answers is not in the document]
drammock: maybe a naive question, but if what you need is the taper weights, why not store those directly rather than storing the other estimation params necessary to recompute them?  It's not a perfect solution because TFRs computed from earlier versions of MNE and then saved to disk may still lack the necessary information... but the same is true of storing the estimation parameters as attributes (instead of the weights).

mne_connectivity/spectral/epochs.py
```diff
@@ -1126,28 +1144,47 @@ def spectral_connectivity_epochs(
...
-                # their weights in the same way as for multitaper spectra
-                mode = "multitaper" if "taper" in data._dims else "fourier"
+            mode = data.method
+            if isinstance(data, EpochsSpectrum | EpochsSpectrumArray):
+                if isinstance(data, EpochsSpectrum):  # read mode from object
+                    mode = "fourier" if mode == "welch" else mode
+                else:  # infer mode from dimensions
+                    # Currently, actual mode doesn't matter as long as we handle tapers
+                    # and their weights in the same way as for multitaper spectra
+                    mode = "multitaper" if "taper" in data._dims else "fourier"
```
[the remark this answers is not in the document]
tsbinns: Good spot! I'll look through the tests again and find where it's slipping through.

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -512,26 +512,33 @@ def test_spectral_connectivity_epochs_spectrum_input(method, mode):
...
     con_from_epochs = spectral_connectivity_epochs(
-        data=data, method=method, indices=indices, mode=mode
+        data=data, method=method, indices=indices, mode=mode, cwt_freqs=cwt_freqs
     )
-    if mode == "multitaper":
-        atol = 0
-    else:
+    # Works for multitaper & Morlet, but Welch of Spectrum and Fourier of spec_conn are
+    # slightly off (max. abs. diff. ~0.006) even when what should be identical settings
+    # are used
```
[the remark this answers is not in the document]
tsbinns: I can't rule out there being some parameter difference I overlooked. I'll try going through both pipelines step-by-step to find where exactly the deviation emerges.

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -546,25 +553,18 @@ def test_spectral_connectivity_epochs_spectrum_input(method, mode):
...
-        con_thresh = (0.1, 0.3)
-    else:  # higher baseline for Welch/Fourier
-        con_thresh = (0.2, 0.4)
-
     # check freqs of simulated interaction show strong connectivity
-    assert_array_less(con_thresh[1], np.abs(con.get_data()[:, freqs_con].mean()))
+    assert_array_less(0.6, np.abs(con.get_data()[:, freqs_con].mean()))
     # check freqs of no simulated interaction (just noise) show weak connectivity
-    assert_array_less(np.abs(con.get_data()[:, freqs_noise].mean()), con_thresh[0])
+    assert_array_less(np.abs(con.get_data()[:, freqs_noise].mean()), 0.3)
```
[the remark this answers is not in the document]
tsbinns: Good point.

mne_connectivity/spectral/time.py
tsbinns: > statements like this worry me, because they mean that it's easy for people to make mistakes.

I 100% agree. With storing the params (or the weights directly like discussed in https://github.com/mne-tools/mne-python/issues/12851), then this statement could be removed and prevent possible mistakes.

> but I still have the feeling like this function is trying to do too much.

But yeah, I also understand.

> an API like this makes me think that there should be separate functions for operating on MNE Objects and on arrays

Just to clarify, this would mean something like:
- `spec_conn_epochs/time` handling `Epochs` and `EpochsSpectrum/TFR` objects
- and a `spec_conn_epochs/time_array` function handling array of timeseries data or coefficients

If there were separate functions, would it not be simpler to maintain `spec_conn_epochs/time` as handling timeseries data (either as `Epochs` or arrays), and introduce equivalent functions for handling spectral coefficients (either as `EpochsSpectrum`, `EpochsTFR`, or arrays)? I feel like this would be more compatible with the suggestion below:

> Another possibility is 2 public-facing funcs and no shared helper; one of the public funcs would then call the other after suitable transformation of its inputs. This could mean internally extracting arrays from objects and always operating on arrays, or it could mean internally promoting arrays to objects and always operating on objects.

However for the second sentence, I think again it's less a case of distinguishing data in arrays vs. MNE objects, and more distinguishing timeseries vs. coefficients.

> Without looking closely I'm not sure how much work that would be / how (in)consistent it is with the rest of the mne_connectivity API.

I imagine the most important thing is for whatever is changed in `spec_conn_time`, that should be mirrored in `spec_conn_epochs` and vice versa (so things like https://github.com/mne-tools/mne-connectivity/pull/220 would need to be looked at again). I am perfectly happy to work on this (but I also understand that means more work for reviewers).

tsbinns:
Thanks for the detailed review @drammock! Until those wider API questions are addressed I'll double check the unit tests and try to find where the deviation in Fourier/Welch pipelines.

mne_connectivity/spectral/time.py
drammock: > I think again it's less a case of distinguishing data in arrays vs. MNE objects, and more distinguishing timeseries vs. coefficients.

yeah you're right I was confusing the issue. Just to make sure we're on the same page: am I right that the "flavors" of spectral connectivity you want to support are:

- input: non-spectral (ordinary time series)
    - across: epochs (`spec_conn_epochs`; objects or arrays)
    - across: times (`spec_conn_time`; objects or arrays)
- input: spectral
    - across: epochs (currently `spec_conn_epochs`; maybe a new func?)
domain)
    - across: time (not supported; no time dimension)
- input: spectrotemporal
    - across: epochs (not supported?)
    - across: times (this PR: `spec_conn_time`; maybe a new func? `spec_conn_tfr_time`?)

mne_connectivity/spectral/time.py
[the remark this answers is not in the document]
tsbinns: Copying from the discussion in https://github.com/mne-tools/mne-python/issues/12851#issuecomment-2360912513 for reference:

> drammock: Since we already store weights for spectrum objects, I'm inclined to store weights (not estimation params) for TFRs too. That saves having to recompute them when users pass MNE objects (not arrays) to the connectivity functions, and is more internally consistent.

>> tsbinns: Sure, I see your reasoning, especially to keep it consistent with *Spectrum objects. I'll open a PR for this.

mne_connectivity/spectral/time.py
tsbinns: > input: non-spectral (ordinary time series)
> - across: epochs (spec_conn_epochs; objects or arrays)
> - across: times (spec_conn_time; objects or arrays)

Yes.

> input: spectral
> - across: epochs (currently spec_conn_epochs; maybe a new func?)
> domain)
> - across: time (not supported; no time dimension)

Yes.

> input: spectrotemporal
> - across: epochs (not supported?)
> - across: times (this PR: spec_conn_time; maybe a new func? spec_conn_tfr_time?)

This PR was also adding support for spectrotemporal input for `spec_conn_epochs` to match the `cwt_morlet` mode.

mne_connectivity/spectral/tests/test_spectral.py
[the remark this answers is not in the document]
tsbinns: So I've gone through both implementations with matching settings (nfft, window func, etc...) and I don't see any bug in the code, instead it seems to just be a difference in how each implementation performs the Fourier transform.

The `psd_array_welch()` function of MNE (called by `Epochs.compute_psd()`) uses `scipy.signal.spectrogram()` for the transformation, while MNE-Connectivity uses `scipy.fft.rfft()`. Running the same data through these functions directly gives very different results in absolute terms due to scaling differences, and when accounting for this while the results become very similar, there is still variation.

Like we see in the tests, this manifests in only some very minor differences for the connectivity results (in this case at most 0.006 for any one frequency bin). Altogether I don't think it's cause for concern, just the result of different Fourier transform pipelines.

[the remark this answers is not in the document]
drammock: OK thanks for the detailed investigation! Can you update the comment saying "difference is due to scipy.signal.spectrogram vs scipy.fft.rfft" or similar?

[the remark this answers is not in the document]
tsbinns: Done!

tsbinns:
Conda test keeps freezing on the micromamba setup step, seems similar to this: https://github.com/mamba-org/setup-micromamba/issues/225
Is it worth trying to fix now?

larsoner:
Yeah feel free! Sounds like there might be a workaround via pinning in some of those issue comments

tsbinns:
Haven't forgotten about this, just need to sort https://github.com/mne-tools/mne-python/pull/12910 before I can finalise the changes and tests here.

tsbinns:
#12910 is now merged, so I can fix the outstanding issues here.

mne_connectivity/spectral/epochs.py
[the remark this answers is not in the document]
tsbinns: Have now updated the order of checks.

mne_connectivity/spectral/epochs.py
```diff
@@ -1126,28 +1144,47 @@ def spectral_connectivity_epochs(
...
+                        raise ValueError(
+                            "if `data` is an EpochsTFR object, the spectral method "
+                            "must be 'morlet'"
+                        )
+                else:
+                    if "taper" in data._dims:  # FIXME: Add support for multitaper TFR
+                        raise ValueError(
+                            "if `data` is an EpochsTFRArray object, it cannot contain "
+                            "Fourier coefficients for individual tapers"
+                        )
```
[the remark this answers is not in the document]
tsbinns: Also updated.

mne_connectivity/spectral/tests/test_spectral.py
tsbinns: Was in anticipation of also checking with TFR data, which has now been added.

mne_connectivity/spectral/time.py
tsbinns: This is now no longer an issue following https://github.com/mne-tools/mne-python/pull/12910

mne_connectivity/spectral/time.py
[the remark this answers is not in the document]
tsbinns: Also addressed in https://github.com/mne-tools/mne-python/pull/12910

mne_connectivity/spectral/epochs.py
```diff
@@ -434,6 +445,37 @@ def _compute_spectra(
...
+    # expand weights dims to match x_mt and y_mt
+    weights_x = np.expand_dims(weights_x, axis=(*np.arange(x_mt.ndim - 3), -1))
+    weights_y = np.expand_dims(weights_y, axis=(*np.arange(y_mt.ndim - 3), -1))
+    # compute CSD
+    csd = np.sum(weights_x * x_mt * (weights_y * y_mt).conj(), axis=-3)
+    denom = np.sqrt((weights_x * weights_x.conj()).real.sum(axis=-3)) * np.sqrt(
+        (weights_y * weights_y.conj()).real.sum(axis=-3)
+    )
+    csd *= 2 / denom
+    return csd
```
tsbinns, line 476: There isn't a `_csd_from_mt` equivalent for TFR data in MNE. Would people be open to me adding this to the main MNE package?
Otherwise it can just remain here.

mne_connectivity/spectral/epochs.py
```diff
@@ -511,14 +554,22 @@ def _epoch_spectral_connectivity(
...
+            if weights is None:  # assumes no tapers dim, i.e., for Fourier/Welch mode
+                x_t = np.expand_dims(x_t, axis=2)  # CSD construction expects tapers dim
+                weights = np.ones((1, 1, 1))  # assign dummy weights
         if accumulate_psd:
-            this_psd = _psd_from_mt(x_t, weights)
+            if weights is not None:  # mode == 'multitaper' or 'fourier'
+                if not is_tfr_con:  # normal spectra (multitaper or Fourier)
+                    this_psd = _psd_from_mt(x_t, weights)
+                else:  # TFR spectra (multitaper)
+                    this_psd = np.array([_tfr_from_mt(epo_x, weights) for epo_x in x_t])
```
tsbinns, line 570: Realising now that the way I implemented `_tfr_from_mt` in https://github.com/mne-tools/mne-python/pull/12910 works fine there, but isn't flexible to an epochs dimension (in contrast `_psd_from_mt` is).

Would people be open to a small PR in the main package to fix this?

mne_connectivity/spectral/time.py
```diff
@@ -66,12 +72,26 @@ def spectral_connectivity_time(
...
+           Fourier coefficients stored in an :class:`~mne.time_frequency.EpochsTFR`
+           object can also be passed in as data. Storing multitaper weights in
+           :class:`~mne.time_frequency.EpochsTFR` objects requires ``mne >= 1.10``.
+    freqs : array_like | None
+        Array of frequencies of interest for time-frequency decomposition. Only the
+        frequencies within the range specified by ``fmin`` and ``fmax`` are used. If
+        ``data`` is an :term:`array-like` or :class:`~mne.Epochs` object, the
+        frequencies must be specified. If ``data`` is an
+        :class:`~mne.time_frequency.EpochsTFR` object, ``data.freqs`` is used and this
+        parameter is ignored.
```
tsbinns, line 94: For TFR data, `freqs` can be read from the object, but in the current state `freqs` must be passed.
If `freqs` could be `None` by default, is this an API change that would require a deprecation cycle?
Of course a check would be made that `freqs != None` when data isn't `EpochsTFR`.

mne_connectivity/spectral/time.py
```diff
@@ -369,12 +392,23 @@ def spectral_connectivity_time(
...
+    _validate_type(
+        data,
+        (np.ndarray, BaseEpochs, EpochsTFR),
+        "`data`",
+        "Epochs, EpochsTFR, or a NumPy array",
+    )
+    if not isinstance(data, EpochsTFR) and freqs is None:
+        raise TypeError(
+            "`freqs` must be specified when `data` is not an EpochsTFR object"
+        )
```
tsbinns, line 408: This is the check mentioned above.

tsbinns:
I've addressed the previous issues.

There should now be full support for `EpochsSpectrum` and `EpochsTFR` data in `spectral_connectivity_epochs` (including time-resolved multitaper coeffs) and for `EpochsTFR` in `spectral_connectivity_time`, along with unit tests covering the changes.

tsbinns:
Given that the multitaper TFR requires the unreleased MNE 1.10, is it worth holding off on merging (once it's checked and ready to go) until after MNE-Conn 0.8 is released?

On that note, since the last major part of the GSoC project in #223 was merged, there's a decent amount for a new release: https://mne.tools/mne-connectivity/dev/whats_new.html
What do people think?

larsoner:
> Given that the multitaper TFR requires the unreleased MNE 1.10, is it worth holding off on merging (once it's checked and ready to go) until after MNE-Conn 0.8 is released?

You could, but even then you'll need to add a version guard to the mne-connectivity that checks the MNE version. You have them in tests but you also need them in the user-facing function. You can use something like `_soft_import("mne", "multitaper connectivity", min_version="1.10")` for it

larsoner:
... although I think the `min_version` in `_soft_import` was added recently, so better just to use
```
if not check_version("mne", "1.10):
    raise ...
```
in the user-facing code

tsbinns:
> You could, but even then you'll need to add a version guard to the mne-connectivity that checks the MNE version. You have them in tests but you also need them in the user-facing function. You can use something like `_soft_import("mne", "multitaper connectivity", min_version="1.10")` for it

Currently I'm checking compatibility based on the object attrs. Do you think an explicit version check is cleaner?

https://github.com/mne-tools/mne-connectivity/blob/73ce0a6bc4ea164656d36af049119090fbd07e06/mne_connectivity/spectral/epochs.py#L1225-L1233

drammock:
> Currently I'm checking compatibility based on the object attrs. Do you think an explicit version check is cleaner?

Is the issue just about which MNE version is currently installed?  Or is it to do with which version of MNE the data were saved to disk with?  If the latter plays a role, then attr-based checking seems right to me (though I'd tweak the error message to explicitly reference the possibility of loaded data that were saved by an older version of MNE).  On the other hand if it's just about the currently installed version of MNE, then checking that (rather than object attrs) makes for easier-to-maintain code IMO (as the code more directly reflects the source of the problem).

tsbinns:
> Is the issue just about which MNE version is currently installed? Or is it to do with which version of MNE the data were saved to disk with?

Moreso the latter, whatever MNE version was used to compute the coefficients.
But thinking about it, could be the case that a multitaper `EpochsSpectrum/TFR` object saved from an older MNE version and loaded with a newer version has a `weights` attr that is `None`, so that check would fail. A more thorough check would be:
```
if not hasattr(data, "weights") or (data.weights is None and mode == "multitaper"):
    # XXX: Remove logic when support for mne<1.10 is dropped 
    raise AttributeError( 
       "weights are required for multitaper coefficients stored in " 
       "EpochsSpectrum (requires mne >= 1.8) and EpochsTFR (requires " 
       "mne >= 1.10) objects" 
    ) 
```

> I'd tweak the error message to explicitly reference the possibility of loaded data that were saved by an older version of MNE

Yeah good call, I'll add this.

drammock (approved):

mne_connectivity/spectral/epochs.py
larsoner: Sure!

mne_connectivity/spectral/time.py
larsoner: I think the `freqs=None` here is okay

mne_connectivity/spectral/epochs.py
larsoner: I think even if you put it there you'd need some sort of `try/except` based on MNE version with a fallback to this code (in this codebase) anyway.

larsoner (approved):

larsoner:
Thanks @tsbinns !

mne_connectivity/spectral/epochs.py
tsbinns: Good point. So we should keep this here for now.

tsbinns:
Thanks @drammock and @larsoner for the patience with this!

PR #241, opened 2024-10-09
[DOC] Use `sphinx_gallery_multi_image` in decomposition example

tsbinns:
Takes advantage of the new `sphinx_gallery_multi_image` option in v0.18 in the decomposition plotting example to make the filters and patterns easier to see.

Before:
image

After:
image


larsoner:
Nice @tsbinns !

PR #242, opened 2024-10-11
[DOC] Improve documentation of freqs

richardkoehler:
Creating this PR after brief discussion with @tsbinns. I can also create a formal issue if needed, but for the moment this is a pure documentation PR.
In brief, I was confused by the behavior of ``cwt_freqs``, ``fmin`` and ``fmax``, especially if both ``cwt_freqs`` plus one of the other parameters was specified. Apparently, ``fmin`` and ``fmax`` supersede the ``cwt_freqs`` parameter, which was not evident from the documentation, so I tried to improve the description to make it a bit clearer. I can add more examples, if you think even more detail is needed.
Additionally, I noticed that the parameter ``None`` of ``fmin``, was not documented, even though it is the default, so I added this, too.
I ran make on the docs and they seem to render fine locally. Let me know if you need anything else.

PR Description
--------------

- Add None as parameter option, as this is actually the default case
- Add description of the behavior of fmin=None
- Add spaces between number and unit Hz (SI style convention)
- Improve description of cwt_freqs behavior for the case when fmin and fmax are also defined by user.

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy

### DIFF
--- a/mne_connectivity/spectral/epochs.py
+++ b/mne_connectivity/spectral/epochs.py
@@ -790,12 +790,15 @@ def spectral_connectivity_epochs(
         Spectrum estimation mode can be either: 'multitaper', 'fourier', or
         'cwt_morlet'. Ignored if ``data`` is an
         :class:`~mne.time_frequency.EpochsSpectrum` object.
-    fmin : float | tuple of float
+    fmin : float | tuple of float | None
         The lower frequency of interest. Multiple bands are defined using
-        a tuple, e.g., (8., 20.) for two bands with 8Hz and 20Hz lower freq.
+        a tuple, e.g., (8., 20.) for two bands with 8 Hz and 20 Hz lower freq.
+        If `None`, the frequency corresponding to 5 cycles based on the epoch
+        length is used. For example, with an epoch length of 1 sec, the lower
+        frequency would be 5 / 1 sec = 5 Hz.
     fmax : float | tuple of float
         The upper frequency of interest. Multiple bands are dedined using
-        a tuple, e.g. (13., 30.) for two band with 13Hz and 30Hz upper freq.
+        a tuple, e.g. (13., 30.) for two band with 13 Hz and 30 Hz upper freq.
     fskip : int
         Omit every "(fskip + 1)-th" frequency bin to decimate in frequency
         domain.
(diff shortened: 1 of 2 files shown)


mne_connectivity/spectral/epochs.py
```diff
@@ -790,12 +790,15 @@ def spectral_connectivity_epochs(
         Spectrum estimation mode can be either: 'multitaper', 'fourier', or
         'cwt_morlet'. Ignored if ``data`` is an
         :class:`~mne.time_frequency.EpochsSpectrum` object.
-    fmin : float | tuple of float
+    fmin : float | tuple of float | None
         The lower frequency of interest. Multiple bands are defined using
-        a tuple, e.g., (8., 20.) for two bands with 8Hz and 20Hz lower freq.
+        a tuple, e.g., (8., 20.) for two bands with 8 Hz and 20 Hz lower freq.
+        If `None`, the frequency corresponding to 5 cycles based on the epoch
```
drammock, line 796:
```suggestion
        If ``None``, the frequency corresponding to 5 cycles based on the epoch
```

drammock (commented):
CI doc build fails with:
```
WARNING: while setting up extension sphinx_issues: role 'cve' is already registered, it will be overridden [app.add_role]
WARNING: while setting up extension sphinx_issues: role 'cwe' is already registered, it will be overridden [app.add_role]
```

I don't think that's related to your changes, but we should still investigate/fix

richardkoehler:
> CI doc build fails with:
> 
> ```
> WARNING: while setting up extension sphinx_issues: role 'cve' is already registered, it will be overridden [app.add_role]
> WARNING: while setting up extension sphinx_issues: role 'cwe' is already registered, it will be overridden [app.add_role]
> ```
> 
> I don't think that's related to your changes, but we should still investigate/fix

I think this might be related to a regression in Sphinx 8.1.0. There was also a bugfix release 8.1.1 related to 'cve' yesterday: https://www.sphinx-doc.org/en/master/changes/index.html#release-8-1-1-released-oct-11-2024. I cannot get the docs to build on my windows machine, so I will try to add a version constraint and see if the tests pass in CI.

drammock (approved):

PR #280, opened 2025-01-27
[BUG] Fix issue in decomposition plotting to specified axes

tsbinns:
When playing around with customising the plotting of the decomposition filters and patterns, I realised the parameter to pass in axes for plotting the data to doesn't work as intended.

Here there is separate plotting for the seeds and targets, but `axes` expects just one set of plots. This means if you pass in the desired axes to plot to, the seeds will be plotted first, then the targets will be plotted and overwrite the seeds.

A solution is to pass in 2 sets of `axes`, one for the seeds and one for the targets. After that everything is handled by the existing topo plotting.


mne_connectivity/utils/docs.py
```diff
@@ -434,14 +434,16 @@
...
 
 docdict["axes_topomap"] = """
-axes : matplotlib.axes.Axes | list of matplotlib.axes.Axes | None (default None)
+axes : list of list of matplotlib.axes.Axes | None (default None)
     The axes to plot to. If `None`, a new figure will be created with the correct number
-    of axes. If not `None`, the number of axes must match ``components``.
+    of axes. If not `None`, there must be two lists containing the axes for the seeds
+    and targets, respectively. In each of these two lists, the number of axes must match
+    ``components`` if ``colorbar=False``, or ``components * 2`` if ``colorbar=True``.
 """
```
tsbinns, line 442: Updated the docstring and also added more info about the required number of axes when colourbar is plotted or not.

mne_connectivity/utils/docs.py
```diff
@@ -434,14 +434,16 @@
...
 
 docdict["name_format_topomap"] = r"""
 name_format : str | None (default None)
-    The string format for axes titles. If `None`, uses ``f"{method}%%01d"``, i.e. the
+    The string format for axes titles. If `None`, uses ``f"{method}%01d"``, i.e. the
     method name followed by the component number.
 """
```
tsbinns, line 448: An unrelated change, but the format suggestion for axes titles had an extra `%`.

doc/whats_new.rst
```diff
@@ -23,9 +23,10 @@ Enhancements
...
 - Add support for computing multiple components of multivariate connectivity in the :func:`~mne_connectivity.spectral_connectivity_epochs` and :func:`~mne_connectivity.spectral_connectivity_time` functions and :class:`~mne_connectivity.decoding.CoherencyDecomposition` class, and add support for storing data with a components dimension in all :class:`~mne_connectivity.Connectivity` classes, by `Thomas Binns`_ and `Eric Larson`_ (:pr:`213`).
 - Add support for :class:`mne.time_frequency.EpochsSpectrum` objects to be passed as data to the :func:`~mne_connectivity.spectral_connectivity_epochs` function, by `Thomas Binns`_ and `Eric Larson`_ (:pr:`220`).
+- Add support for :class:`mne.time_frequency.EpochsTFR` objects to be passed as data to the :func:`~mne_connectivity.spectral_connectivity_epochs` and :func:`~mne_connectivity.spectral_connectivity_time` functions, by `Thomas Binns`_ and `Daniel McCloy`_ (:pr:`232`).
```
tsbinns, line 29: This was missing before #232 was merged. My bad.

tsbinns:
Would also like to get opinions on how figure `suptitle` is handled.

By default, the seeds figure gets the title `"Seeds"`, then the targets figure `"Targets"`. However, if custom `axes` as passed corresponding to the same figure, the seed title is set first, which is then overwritten with the targets title.

This can be very easily fixed after the fact with a call to `fig.suptitle()` and the plotting functions do already have a very lengthy set of params, but it could be cleaner to have a `title` param of some sort. What do people think?

mne_connectivity/utils/docs.py
drammock: was this actually rendering wrong? I think we need an extra `%` if we want a literal `%` sign in docstrings that use `fill_doc` or `verbose`

mne_connectivity/utils/docs.py
```diff
@@ -434,14 +434,16 @@
...
 
 docdict["axes_topomap"] = """
-axes : matplotlib.axes.Axes | list of matplotlib.axes.Axes | None (default None)
+axes : list of list of matplotlib.axes.Axes | None (default None)
```
drammock, line 437: I don't love "list of list of X" as an API description. A naive user might assume a typo, and even a savvy user might just gloss over the duplicated words when reading quickly. Options:

- say "length-2 tuple of lists of matplotlib.axes.Axes" or something like that. Even if we allow list-of-list, saying "tuple-of-lists" is easier to understand IMO
- deprecate `axes` and introduce 2 new params, `seed_axes` and `target_axes`.

Open to other ideas if you have them.

doc/whats_new.rst
```diff
@@ -23,9 +23,10 @@ Enhancements
 ~~~~~~~~~~~~
 
 - Add a new :class:`~mne_connectivity.decoding.CoherencyDecomposition` class for decomposing connectivity sources using multivariate coherency-based methods, by `Thomas Binns`_ (:pr:`193`).
-- Add new plotting methods :meth:`CoherencyDecomposition.plot_filters() <mne_connectivity.decoding.CoherencyDecomposition.plot_filters>` and :meth:`CoherencyDecomposition.plot_patterns() <mne_connectivity.decoding.CoherencyDecomposition.plot_patterns>` for visualising the decomposed connectivity sources, by `Thomas Binns`_ (:pr:`208`).
+- Add new plotting methods :meth:`CoherencyDecomposition.plot_filters() <mne_connectivity.decoding.CoherencyDecomposition.plot_filters>` and :meth:`CoherencyDecomposition.plot_patterns() <mne_connectivity.decoding.CoherencyDecomposition.plot_patterns>` for visualising the decomposed connectivity sources, by `Thomas Binns`_ (:pr:`208`) and (:pr:`279`).
```
tsbinns, line 26: Added the changelog entry for this PR alongside when the plotting functions were first added. Not sure if this is the best way to handle.
Should I instead add a bug fix entry even if the plotting feature hasn't been released yet?

drammock: Fine to handle it this way (better, IMO, than adding a bugfix entry, since this is within-release).  But put both PRs within the same set of parentheses please, like `(:pr:208 and :pr:209)`

mne_connectivity/utils/docs.py
tsbinns: I thought that too, but seems it renders like it's written in the code which doesn't work properly in the plots:
image

doc/whats_new.rst
tsbinns: Fixed!

drammock:
> Would also like to get opinions on how figure `suptitle` is handled.
> 
> By default, the seeds figure gets the title `"Seeds"`, then the targets figure `"Targets"`. However, if custom `axes` as passed corresponding to the same figure, the seed title is set first, which is then overwritten with the targets title.
> 
> This can be very easily fixed after the fact with a call to `fig.suptitle()` and the plotting functions do already have a very lengthy set of params, but it could be cleaner to have a `title` param of some sort. What do people think?

In MNE-Python I've generally been moving *away* from including `title` params (because `fig.suptitle` is so easy to do), except in cases where the automatic title includes some informative / calculated value. For example, funcs that generate 3 topoplot figures (mag, grad, EEG) need the title to differentiate (it would be *possible* to infer which is which from the sensor layouts and the fact that grad is strictly positive... but it's a pain).

In this case, off the cuff I'd be inclined to auto-title each *axes* as "seed 0", "seed 1", ..., "target 0", "target 1", etc. and *not* include an automatic figure title. WDYT?

mne_connectivity/utils/docs.py
drammock: ah OK then it should be removed. 👍🏻

mne_connectivity/utils/docs.py
tsbinns: Sure, I see what you mean. Went for your first suggestion.

tsbinns:
> In MNE-Python I've generally been moving away from including title params (because fig.suptitle is so easy to do), except in cases where the automatic title includes some informative / calculated value. For example, funcs that generate 3 topoplot figures (mag, grad, EEG) need the title to differentiate (it would be possible to infer which is which from the sensor layouts and the fact that grad is strictly positive... but it's a pain).
> In this case, off the cuff I'd be inclined to auto-title each axes as "seed 0", "seed 1", ..., "target 0", "target 1", etc. and not include an automatic figure title. WDYT?

Sure, that makes sense.

Just pushed some changes to remove the `suptitle` and added the seeds/targets name to the axes titles.
I also added a little more info in the docstring that a format specifier for the component number is required (this is expected by the topomap plotting).

Pretty happy with this:
image

mne_connectivity/utils/docs.py
```diff
@@ -443,8 +443,10 @@
 
 docdict["name_format_topomap"] = r"""
 name_format : str | None (default None)
-    The string format for axes titles. If `None`, uses ``f"{method}%01d"``, i.e. the
-    method name followed by the component number.
+    The string format for axes titles. If `None`, uses ``f"{method}%01d_{group}"``,
```
drammock, line 446:
```suggestion
    The string format for axes titles. If ``None``, uses ``f"{method}%01d_{group}"``,
```

drammock (approved):
This looks good.  If I were to nitpick, I would say 2 things (both of which you can take or leave / use your own judgment):

1. The axes titles don't need underscores, just use spaces (and maybe parentheses?) e.g. "MIC 0 (seeds)"
2. It's a bit odd that the `name_format` specifier mixes f-string and %-string formatting.  It could be something like `f"{method} {idx:d} ({group})"` (to yield what I suggest above in (1)). You'd need to update the docstring to explain what `idx` is of course.

tsbinns:
> The axes titles don't need underscores, just use spaces (and maybe parentheses?) e.g. "MIC 0 (seeds)"

Sure, have pushed this change.

tsbinns:
> It's a bit odd that the name_format specifier mixes f-string and %-string formatting. It could be something like f"{method} {idx:d} ({group})" (to yield what I suggest above in (1)). You'd need to update the docstring to explain what idx is of course.

The %-string formatting for the component numbers is used by [`plot_evoked_topomap()`](https://github.com/mne-tools/mne-python/blob/715540a823ae5dec335bee0b2499f1f7183c19c4/mne/viz/topomap.py#L2361) (there considered as time points), so this would need to be changed in the main MNE package.

drammock (approved):

mne_connectivity/utils/docs.py
drammock: sorry, I didn't spot this question before ticking auto-merge. Here's an answer anyway:

In the context of this sentence, `None` is Python code, in the sense of "if you pass `None` here's what you'll get" (substitute `None` for `True` or `"auto"` and it's clear).  That's why I would double-backtick it. Now, in rST, single-backticks are annoying because they mean "try to resolve as a cross-reference using the default domain (which is `:py:obj:` for us I think), but if you can't resolve the xref, just treat like single-asterisks (AKA, italicize)".  Our (probably not written down!) habit/tendency/policy is to never use single backticks (because of this ambiguity) and to always use explicit `:func:` `:class:` `:meth:` etc roles when we want an xref, always use double-backticks when we want code, and always use asterisks when we want italics.

PR #286, opened 2025-02-10
Change calls to `_tfr_from_mt` with support for ndarrays

tsbinns:
Fixes #285.

`_tfr_from_mt` now supports ndarray after https://github.com/mne-tools/mne-python/pull/13104. Means we can avoid looping over epochs dimensions when passing in data.

Also fixes a bug from #232 where `_tfr_from_mt` would be imported even when not required, making everything incompatible with MNE<1.10. Now it is only imported when required.

PR #303, opened 2025-05-23
[MAINT] Run CIs on MNE-stable and update to latest Python version

tsbinns:
Suggested by @larsoner. Currently only run unit tests on `mne-main` which can lead to bugs slipping through (e.g. #286).

This adds a test for the stable MNE version on Ubuntu. Should it be tested on more platforms like in [MNE-BIDS](https://github.com/mne-tools/mne-bids/blob/4f12d190d19a201a5b74fa1c8aaedac4d87775c8/.github/workflows/unit_tests.yml#L145-L147)?

Also updates the latest Python version tested from 3.12 to 3.13.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -66,7 +66,6 @@ jobs:
         if: matrix.mne-version == 'mne-main'
         run: pip install git+https://github.com/mne-tools/mne-python@main
       - run: python -c "import mne; print(mne.datasets.testing.data_path(verbose=True))"
-        if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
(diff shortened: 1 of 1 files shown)


tsbinns:
Also, should this be updated to a Python version we're actually testing or just remove it?
https://github.com/mne-tools/mne-connectivity/blob/bccdcebd75adcff0850a2fbc702f68221e28b96f/.github/workflows/unit_tests.yml#L68-L69

larsoner:
> Also, should this be updated to a Python version we're actually testing or just remove it?

Ouch yes it means we've accidentally been skipping some tests I think. I would actually just get rid of the conditional, I think we always want the testing dataset

larsoner (approved):

Issue #301, opened 2025-05-14
Add the wSMI (weighted Symbolic Mutual Information) marker computation function to the library


Laouen:
### Describe the new feature or enhancement

The wSMI [1] is a connectivity measure developed by Jean-Rémi King and Jacobo D. Sitt and its used mainly to study markers of consciousness and conscious perception and its somehow complementary with the wPLI measure.

The wSMI is already been used by multiple researchs and there are a widely number of publications already including this measure in their projects using either the implementation of Federico Raimondo in PICNIC Lab (where the measure was developed) or their own implementations. We think that adding the measure to MNE instead of everyone using a different implementation would improve the reproducibility of the publications using the measure.  

[1] King, J. R., Sitt, J. D., Faugeras, F., Rohaut, B., El Karoui, I., Cohen, L., ... & Dehaene, S. (2013). Information sharing in the brain indexes consciousness in noncommunicative patients. Current biology, 23(19), 1914-1919.

### Describe your proposed implementation

wSMI is already publicly available in github (https://github.com/fraimondo/nice/blob/master/nice/measures/connectivity.py) developed by Federico Raimondo in python. Even do the measure is public, its not easy to use and installing all the repository for that measure is an overkill. Also, there is no documentation and its not correctly maintained.

Currently we are working on a refactor of our pipeline and we would be available to contribute with MNE integrating the wSMI (which is already implemented) into MNE ourselves.   

### Describe possible alternatives

We don't have a possible alternative but we are open to any

larsoner:
@tsbinns any thoughts on adding this measure?

Laouen:
Thanks for the transfer! Just to specif, when I say we I mean "Picnic lab". I'm currently a PhD student at that lab under Jacobo sitt supervision> I think this makes more sense for mne-connectivity so I'll transfer there!

tsbinns:
Sorry for the delay on this, had some trouble getting the old code working, so I definitely agree with @Laouen that this existing implementation is not so practical to use. In terms of the measure itself, I see the benefits.

For an implementation in MNE-Connectivity, I imagine a standalone function like for `envelope_correlation` or `phase_slope_index` is most appropriate (definitely not a fit for `spec_conn_epochs/times`). Would that be okay @larsoner?

larsoner:
Yeah a new function makes sense if needed

tsbinns:
[...]

PR #307, opened 2025-06-17
ENH: Add weighted symbolic mutual information (wSMI) connectivity measure

GioMarraffini:
Thanks for contributing. If this is your first time,
make sure to read [contributing.md](https://github.com/mne-tools/mne-connectivity/blob/main/CONTRIBUTING.md)

PR Description
--------------

Add wsmi() function to compute weighted symbolic mutual information, a connectivity measure for studying markers of consciousness and conscious perception based on symbolic dynamics.

The implementation includes: - Symbolic transformation using ordinal patterns with optimizations - Mutual information computation between symbolic sequences using Numba JIT - Weighting based on pattern distance for enhanced sensitivity - Memory checks for large kernel sizes to prevent excessive memory usage - Comprehensive test suite with ground truth validation against reference data - Integration with MNE-Connectivity's EpochTemporalConnectivity class - Support for Current Source Density (CSD) preprocessing for EEG data - Configurable low-pass filtering with automatic frequency selection

The measure quantifies non-linear statistical dependencies between time series and is particularly useful for studying consciousness-related brain dynamics.
This PR closes <#301> (Issue number)

Merge checklist
---------------

- [ ] All comments resolved
- [ ] This is not your own PR
- [ ] All CIs are happy


mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+
+        epochs_csd = compute_current_source_density(epochs_temp, lambda2=1e-5)
+        # Check if CSD actually produced CSD channels
+        if pick_types(epochs_csd.info, csd=True).size > 0:
+            epochs = epochs_csd
+        else:
+            logger.warning(
+                """CSD computation did not result in any CSD channels.
+                Using original EEG data for EEG channels."""
+            )
```
tsbinns, line 332: I'm not sure this should be in the connectivity estimation function, I feel it is overburdening it a bit. My instinct would be to leave this to each user to implement in their own analysis pipeline.
For the accompanying example, applying the CSD could be something shown there as an optional preprocessing step.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+        ref_meg=False,
+        exclude="bads",
+    )
+
+    if len(picks) == 0:
+        raise ValueError(
+            """No suitable channels (MEG, EEG, CSD, SEEG, ECoG)
+            found after picking logic.
+            Check channel types and 'bads'."""
+        )
```
tsbinns, line 352: I don't think it is necessary to specify these data channel types. We don't do this elsewhere in `MNE-Connectivity`, so just calling `get_data()` should be enough for selecting the data channels.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+    filter_freq=None,
+    csd=True,
+    weighted=True,
+    memory_limit_gb=1.0,
+    verbose=None,
+):
+    """Compute weighted symbolic mutual information (wSMI).
+
+    Parameters
+    ----------
```
tsbinns, line 240: Should an `indices` parameter for selecting the channels of interest be present? That is the typical API of `MNE-Connectivity`.
Is there something about the WSMI method that would complicate this and require us to always compute it on all channels?

Also, I think it would be good to add an `average` parameter for averaging across epochs, like in `spectral_connectivity_time()`.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+        Must be > 1. Values > 7 may require significant memory.
+    tau : int
+        Time delay (lag) between consecutive pattern elements.
+        Must be > 0.
+    tmin : float | None
+        Time to start connectivity estimation. If None, uses beginning
+        of epoch.
+    tmax : float | None
+        Time to end connectivity estimation. If None, uses end of epoch.
+    filter_freq : float | None
```
tsbinns, line 254: Does this assume the data has already been high-passed? Elsewhere in `MNE-Connectivity` there are parameters for controlling both the min and max frequencies, so I wonder if there should also be a high-pass filter parameter (that by default does nothing).

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+                result_conn_data[epoch_idx, conn_idx, 0] = result_epoched[
+                    epoch_idx, i, j
+                ]
+    else:
+        # For single channel or no channels, create empty connectivity
+        logger.info(
+            f"Only 1 channel selected, {method_name} connectivity will be empty."
+        )
+        result_conn_data = np.empty((n_epochs, 0, 1))
+        indices_list = []
```
tsbinns, line 457: I would expect this to raise an error rather than returning an empty connectivity object. Perhaps also with an `indices` parameter problematic cases like this could be caught early?

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,967 @@
...
+
+        # Compare results
+        expected_data = expected_output.get_data()
+        new_data = conn.get_data()
+
+        assert_allclose(expected_data, new_data, rtol=1e-10, atol=1e-10)
+
+
+def test_wsmi_performance_comparison():
+    """Compare performance of new implementation against the test data timing.
```
tsbinns, line 848: We don't check against performance benchmarks elsewhere in `MNE-Connectivity`, so I wonder if there are some scenarios in which this could cause tests to fail despite the output values being correct.

@larsoner Is there a precedent for this in `MNE` we should follow?

tsbinns (commented):
Thanks for opening the PR! Below are a couple suggestions/comments.

Typically, new features like this would also include an example. This can be just a very short demonstration of using the function (e.g., https://mne.tools/mne-connectivity/dev/auto_examples/mne_inverse_envelope_correlation_volume.html) or it could be a more detailed tutorial with an explanation of the method (e.g., https://mne.tools/mne-connectivity/dev/auto_examples/dpli_wpli_pli.html).

I haven't checked the unit tests extensively yet, will revisit based on the discussions. From a first glance they look extensive and the regression tests are great to have.

mne_connectivity/wsmi.py
Laouen: I agree with this point, we had it like this in our version, but the more generic one allows the user to pre select whatever their want and that is more flexible!

pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 dependencies = [
   'mne >= 1.6',
   'netCDF4 >= 1.6.5',
+  'numba >= 0.55.0',
```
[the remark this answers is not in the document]
Laouen: We use numba as an optimization, we could implement it without numbe using multiprocessing, but if numba dependency is not a problem then we would prefer this.

mne_connectivity/wsmi.py
Laouen: no, there is no problem on selecting channels. we can add the parameter

mne_connectivity/wsmi.py
Laouen: well this is a discussion point. wSMI has a parameter tau that in combination with the sampling rate of the signal it will compute the wSMI over different frequency bands, this is because tau specify the distance (in timepoints) between consecutive points of the symbols. Then when computing the signal to symbol transformation, depending on tau and the sampling frecuency, the symbol will represent patterns of the signal over different frequency bands. 

We do a low pass filter amplify this effect, but we will discuss this better to see if it make sense to move to a min-max frequency filter given by the user or not and we will come back to you

GioMarraffini: After further discussion, we've made a design change that better reflects the function's core purpose. We removed the filter_freq parameter entirely and replaced it with an anti_aliasing parameter (default True).

The filtering isn't for general frequency band selection—it's specifically to prevent aliasing artifacts in the symbolic transformation.
Mathematically motivated: When tau > 1, the effective sampling rate becomes sfreq/tau. Without anti-aliasing, spectral content above sfreq/(2*tau) will alias and corrupt ordinal patterns.

Users can't accidentally create temporal-spectral mismatches like analyzing fast gamma rhythms (30-80 Hz) with slow temporal sampling (tau=10).

Advanced users who want full preprocessing control can set anti_aliasing=False (with appropriate warnings).

The core insight: 
Unlike spectral connectivity methods where users select arbitrary frequency bands, wSMI's temporal parameters (tau, kernel) fundamentally determine which frequencies are meaningful. The automatic anti-aliasing frequency sfreq/(kernel*tau) ensures the spectral content matches the temporal scale being analyzed.

User workflow:
Want alpha band connectivity? Filter your epochs first, then call wsmi()
Want slow network dynamics? Use larger tau and let anti-aliasing handle the rest
Want full control? Use anti_aliasing=False after proper preprocessing

What do you think? we have just made a new commit with this changes

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+    epochs : ~mne.Epochs
+        The data from which to compute connectivity.
+    kernel : int
+        Pattern length (symbol dimension) for symbolic analysis.
+        Must be > 1. Values > 7 may require significant memory.
+    tau : int
+        Time delay (lag) between consecutive pattern elements.
+        Must be > 0.
+    tmin : float | None
+        Time to start connectivity estimation. If None, uses beginning
```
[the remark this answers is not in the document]
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+        Increase if you have more RAM and want to use larger kernels.
+    %(verbose)s
+
+    Returns
+    -------
+    conn : instance of EpochTemporalConnectivity
+        Computed connectivity measures. The connectivity object contains
+        the weighted symbolic mutual information values between all channel pairs
+        (if weighted=True) or the standard symbolic mutual information values
+        between all channel pairs (if weighted=False).
```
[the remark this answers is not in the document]
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+    filter_freq : float | None
+        Low-pass filter frequency in Hz. If None, defaults to
+        sfreq / (kernel * tau).
+    csd : bool
+        Whether to apply Current Source Density (CSD) computation
+        for EEG channels. Default is True.
+    weighted : bool
+        Whether to compute weighted SMI (wSMI) or standard SMI.
+        If True (default), computes wSMI with distance-based weights.
+        If False, computes standard SMI without weights.
```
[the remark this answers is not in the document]
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
GioMarraffini: Done! I've changed it so it no longer handles preprocessing. Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

mne_connectivity/wsmi.py
GioMarraffini: Done! Please resolve conversation, or let me know if anything’s unclear.

tsbinns:
@GioMarraffini @Laouen Thanks both for responding to my initial suggestion so quickly. The changes with the anti-aliasing parameter look very reasonable.
Unfortunately the next 2 weeks at work are super busy for me so I will likely be unable to properly review these latest changes before then, but after that I will try to respond ASAP!

GioMarraffini:
> @GioMarraffini @Laouen Thanks both for responding to my initial suggestion so quickly. The changes with the anti-aliasing parameter look very reasonable. Unfortunately the next 2 weeks at work are super busy for me so I will likely be unable to properly review these latest changes before then, but after that I will try to respond ASAP!

@tsbinns Hi just wanted to ask if you had time to review it and if there were any news. We are happy to make any necessary changes.

tsbinns (commented):
Sorry for the delay. Altogether I think the changes were very good and I trust your judgement for adapting the anti-aliasing parameter. Here are a few more comments from me.

tsbinns:
Rerunning the failing test; should not be a problem from your end.

GioMarraffini:
@tsbinns Thanks for the revisions and additional comments! I've just pushed the new updates. let me know if you spot anything else.

mne_connectivity/wsmi.py
```diff
@@ -0,0 +1,483 @@
...
+    weighted=True,
+    memory_limit_gb=1.0,
+    verbose=None,
+):
+    """Compute weighted symbolic mutual information (wSMI).
+
+    Parameters
+    ----------
+    epochs : ~mne.Epochs
+        The data from which to compute connectivity.
```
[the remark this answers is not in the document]
drammock: agree that if other connectivity est funcs support both MNE objects and NumPy arrays, that this one should do likewise.

pyproject.toml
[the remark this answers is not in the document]
drammock: in MNE-Python, Numba is an optional dependency, and we have some triage code that makes the `@jit` decorator a pass-through if numba can't actually be imported:

https://github.com/mne-tools/mne-python/blob/1c9ac7c145f55416661a912a1de66302d2433ba6/mne/fixes.py#L622-L631

If the speedup is huge for these particular calculations, it's probably worth doing something similar here in MNE-Connectivity.

mne_connectivity/tests/test_wsmi.py
tsbinns: In that case @GioMarraffini @Laouen I would suggest removing the benchmark tests and the log can be used to evaluate the performance ocassionally.

tsbinns:
Something I also didn't consider before is the size of the testing data. The existing test data is 0.9 MB, but the wSMI data is 5.3 MB. Without the wSMI test data the source directory is only ~3 MB.
Would it be possible to reduce the size of the test data? Or could some of this be offloaded to simulating the data in the tests at runtime?

tsbinns:
> The problem with generating test data online is that we have no ground thruth for that data. Currently, the ground truth is the value obtained by the original function.

So the way we have handled this in the past when porting methods from other toolboxes is to have a regression test involving some arbitrary connectivity data of only a few channels/epochs and comparing the results from the original and our implementations. This way the data we need to save is very small.
- Example of the data: https://github.com/mne-tools/mne-connectivity/tree/main/mne_connectivity/spectral/tests/data
- Example of the test: https://github.com/mne-tools/mne-connectivity/blob/f6b2b191fea6d6ed7129798455fc832dd033e7a7/mne_connectivity/spectral/tests/test_spectral.py#L870-L921

> Maybe we can do some tests with estimated result patterns and if we see those partterns then assume is correct even do we don't compare with exact results.

This is what we have done for evaluating whether ported methods are working as intended. We simulate test data in the scripts with various connectivity profiles and validate the results based on whether the connectivity values are in a reasonable range. We don't have to check that they match the original implementation exactly since the regression test is handling that side of things.
- Example of testing that connectivity is in a reasonable range:
https://github.com/mne-tools/mne-connectivity/blob/f6b2b191fea6d6ed7129798455fc832dd033e7a7/mne_connectivity/spectral/tests/test_spectral.py#L692-L702

examples/wsmi.py
```diff
@@ -0,0 +1,455 @@
...
+raw_fname = data_path / "MEG" / "sample" / "sample_audvis_filt-0-40_raw.fif"
+event_fname = data_path / "MEG" / "sample" / "sample_audvis_filt-0-40_raw-eve.fif"
+
+# Load and preprocess data
+raw = mne.io.read_raw_fif(raw_fname, verbose=False)
+events = mne.read_events(event_fname, verbose=False)
+
+# Pick a few EEG channels for demonstration
+eeg_picks = ["EEG 001", "EEG 002", "EEG 003", "EEG 004"]
+raw.pick_channels(eeg_picks)
```
tsbinns, line 302: Something I just noticed: `pick_channels()` is deprecated so let's use the recommended `pick()`.

```suggestion
raw.pick(eeg_picks)
```

tsbinns (commented):
Hi @GioMarraffini, overall the changes were great and adressed the previous comments. There are just a few things regarding the array inputs.

Also, we don't get notifications for commits, so just add a comment or ping us once you think the code is ready for review.

mne_connectivity/wsmi.py
```diff
@@ -324,21 +333,20 @@ def wsmi(
...
-                "Array input detected with default sfreq=1.0 Hz. "
-                "Anti-aliasing filter frequency will be computed based on this. "
-                "Consider using MNE Epochs object for proper frequency handling.",
-                UserWarning,
-                stacklevel=2,
-            )
+
+        # Handle names parameter
+        if names is None:
+            ch_names = [f"CH_{i}" for i in range(n_channels)]
```
tsbinns, line 342: I would suggest not specifying names to use and just allow `None` to be passed through to the connectivity object at the end. That way behaviour of automatic channel names will match that of other functions like `spectral_connectivity_epochs()`, `phase_slope_index()`, `envelope_correlation()`, etc...

mne_connectivity/tests/test_wsmi.py
```diff
@@ -1011,3 +1007,64 @@ def test_wsmi_connectivity_patterns():
...
+    data_matrix = result_default.get_data()
+
+    # We expect a lower-triangular connectivity matrix
+    n_connections = n_channels * (n_channels - 1) // 2
+    assert data_matrix.shape == (n_epochs, n_connections)
+    assert np.all(np.isfinite(data_matrix))
+
+    # Check that default channel names were used
+    expected_names = [f"CH_{i}" for i in range(n_channels)]
+    assert result_default.names == expected_names
```
tsbinns, line 1034: Also means you don't have to check default channel names because that isn't being handled by the function anymore.

mne_connectivity/tests/test_wsmi.py
```diff
@@ -1011,3 +1007,64 @@ def test_wsmi_connectivity_patterns():
...
+    n_epochs, n_channels, n_times = 3, 4, 200
+    rng = np.random.RandomState(42)
+    data = rng.randn(n_epochs, n_channels, n_times)
+
+    # Test with default names
+    result_default = wsmi(data, kernel=3, tau=1, sfreq=sfreq)
+
+    # Check the result
+    assert hasattr(result_default, "get_data")
+    data_matrix = result_default.get_data()
```
tsbinns, line 1025: Nitpick, but not sure the assert is really required since if the returned object doesn't have this method the failure of the test will be clear enough on its own.

tsbinns (commented):
@GioMarraffini Just some very minor things again.

tsbinns:
> Hi @tsbinns , I've addressed your latest feedback. The implementation now follows the pattern you suggested. Ready for review!

Looks good!

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,1065 @@
...
+    epochs = EpochsArray(data, info, tmin=0.0)
+
+    # Test with MEG data
+    conn = wsmi(epochs, kernel=3, tau=1)
+
+    assert conn.n_nodes == n_channels
+    assert conn.method == "wSMI"
+
+
+def test_wsmi_all_channels_as_bad():
```
drammock, line 327: you're already testing what happens if 1 channel is bad, and since the behavior is apparently to ignore bad channels, we shouldn't bother also testing when 2, 3, 4, or all channels are bad.

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,1065 @@
...
+    # Test different kernel and tau combinations
+    conn = wsmi(
+        epochs,
+        kernel=kernel,
+        tau=tau,
+    )
+
+    assert conn.method == "wSMI"
+    assert conn.n_nodes == n_channels
+    assert np.all(np.isfinite(conn.get_data()))
```
drammock, line 390: A test that parametrizes over `kernel` and `tau` values should test that the different `kernel` or `tau` values have different effects. There are at least 2 ways to do that:

1. use non-random channel data, compute from first principles what the conn values should be for different kernel/tau values, pass them in as part of the parametrization, and assert that you get what you expect
2. don't parametrize, run the wsmi a few times (with different kernel/tau values) within the test function, and assert that the resulting `conn.get_data()` arrays are actually different.

drammock (changes requested):
will resume review next week.

pyproject.toml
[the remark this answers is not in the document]
tsbinns: Good catch, yes this should be removed.

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,1065 @@
...
+
+    # Mark one channel as bad
+    epochs.info["bads"] = ["Pz"]
+
+    # Test with bad channels
+    result = wsmi(epochs, kernel=3, tau=1)
+
+    # Check the result - should process all channels (including bad ones)
+    data_matrix = result.get_data()
+    # All 4 channels are processed (bad channels are not excluded)
```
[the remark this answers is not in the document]
drammock: sorry, my question was not clear.  I wasn't commenting on `get_data()` but on `wsmi()` --- IIUC it will ignore bad channel marking and just process all channels. That is what I find surprising (why compute connectivity between good & bad channels, or between 2 channels both marked as bad? Seems a waste of compute resources).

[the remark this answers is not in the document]
tsbinns: Sorry, I just meant to explain that if `indices=None`, everything returned from `get_data()` is used, which includes bad channels. I agree it is inefficient and not so in line with how bads are handled elsewhere in MNE.

Are people in favour of changing the behaviour so that `indices=None` excludes bad channels from being analysed? If so, I could open a PR that introduces a warning that behaviour of bads with `indices=None` will change in (what would be the subsequent release of) v0.9?

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,1065 @@
...
+    assert np.all(np.isfinite(conn_without_filter.get_data()))
+
+    # Results should be different due to filtering
+    # (though with random data, differences might be subtle)
+    assert conn_with_filter.get_data().shape == conn_without_filter.get_data().shape
+
+
+# =============================================================================
+# Ground Truth Validation Tests (New Test Data System)
+# =============================================================================
```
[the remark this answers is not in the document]
GioMarraffini: Hi @drammock ! thanks for all the feedback! just one question should I wait until these following reviews on this file or should I start deleting previous tests?  I already made the necesary adjustment to the example file. Please let me know if it needs further modifications. Thanks!

mne_connectivity/tests/test_wsmi.py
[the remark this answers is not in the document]
GioMarraffini: Should we maybe proceed for now in this PR as it is (using everything returned from `get_data()` ) and change this behaviour in this next following PR? Please let me know how should I proceed. Thanks

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,1065 @@
...
+    ch_names = ["Fz", "Cz", "Pz", "Oz"]
+    info = create_info(ch_names, sfreq=sfreq, ch_types="eeg")
+    epochs = EpochsArray(data, info, tmin=0.0)
+
+    # Set a standard montage
+    montage = mne.channels.make_standard_montage("standard_1020")
+    epochs.set_montage(montage, match_case=False)
+
+    # Mark one channel as bad
+    epochs.info["bads"] = ["Pz"]
```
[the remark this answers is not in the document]
tsbinns: Discussed in meeting that a fixture may not be necessary following changes to the tests, but I will keep this in mind.

mne_connectivity/tests/test_wsmi.py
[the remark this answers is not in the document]
drammock: Yes feel free to start simplifying the tests now. I'll try to take another look on Friday

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+        "wSMI should be identical for identical source channels"
+    )
+
+    # 5. Reasonable bounds (wSMI should not exceed theoretical maximum)
+    assert np.all(avg_conn < 2.0), f"wSMI values too high: max={np.max(avg_conn):.3f}"
+
+    # 6. Test reproducibility across epochs
+    assert np.all(np.std(conn_data, axis=0) < 0.5), (
+        "wSMI should be stable across epochs"
+    )
```
drammock: TODO check stringency is appropriate

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+def test_wsmi_tau_nonlinear_detection():
+    """Test that higher tau values better detect nonlinear coupling."""
+    sfreq = 100.0
+    n_epochs, n_times = 5, 400
+    t = np.linspace(0, n_times / sfreq, n_times)
+
+    # Create test data with clear nonlinear coupling
+    data = np.zeros((n_epochs, 3, n_times))
+
+    for epoch in range(n_epochs):
```
drammock: should be able to simulate without a for-loop?

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+        )
+
+    with open(test_data_path, "rb") as f:
+        test_data = pickle.load(f)
+
+    return test_data
+
+
+def test_wsmi_ground_truth_validation():
+    """Test wSMI against ground truth data for regression testing."""
```
drammock: mention here where the ground truth data comes from (IIUC it's the original implementation provided with the paper?)

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+    conn_smi = wsmi(epochs, kernel=3, tau=1, weighted=False)
+
+    # Should have different method names
+    assert conn_wsmi.method == "wSMI"
+    assert conn_smi.method == "SMI"
+
+    # Should produce different results due to weighting
+    assert not np.array_equal(conn_wsmi.get_data(), conn_smi.get_data()), (
+        "wSMI and SMI should produce different results"
+    )
```
drammock: again, just checking that "results are different" is not stringent enough. At minimum it should be possible to say "one is larger than the other".  Ideally, with known inputs, it should also be possible to say *by how much* it should be larger.

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+    assert conn.n_epochs_used == 2
+
+    # Check data shape and validity
+    data_matrix = conn.get_data()
+    expected_connections = 4 * 3 // 2  # 4 choose 2 = 6 connections
+    assert data_matrix.shape == (2, expected_connections)
+    assert np.all(np.isfinite(data_matrix))
+
+    # Test with different channel types (suppress unit change warning)
+    import warnings
```
drammock: import should be at top of file...  but don't do it that way either.  Use `with pytest.warns(...)`

mne_connectivity/tests/test_wsmi.py
```diff
@@ -0,0 +1,582 @@
...
+
+        # Compare with expected output
+        expected_data = test_case["expected_output"].get_data()
+        new_data = conn.get_data()
+
+        assert expected_data.shape == new_data.shape
+        assert_allclose(expected_data, new_data, rtol=1e-10, atol=1e-10)
+
+
+def test_wsmi_bad_channels():
```
drammock: parametrize to coalesce this test and the following test.

tsbinns: Now addressed

tsbinns:
@GioMarraffini Thanks for the detailed answer! That makes sense.

I think it would be good to include a short note about that in the example, perhaps after the results are shown in the [Visualizing wSMI Connectivity Results](https://output.circle-artifacts.com/output/job/1a6e67e6-aee2-491d-a825-c3a54c453c4a/artifacts/0/dev/auto_examples/wsmi.html#visualizing-wsmi-connectivity-results) section.

Could be something like:
> **Note:** Small negative values can occur for wSMI, reflecting estimator noise or a lack of information sharing above chance, rather than an anti-coupling of signals.

What do you think?

GioMarraffini:
> @GioMarraffini Thanks for the detailed answer! That makes sense.
> 
> I think it would be good to include a short note about that in the example, perhaps after the results are shown in the [Visualizing wSMI Connectivity Results](https://output.circle-artifacts.com/output/job/1a6e67e6-aee2-491d-a825-c3a54c453c4a/artifacts/0/dev/auto_examples/wsmi.html#visualizing-wsmi-connectivity-results) section.
> 
> Could be something like:
> 
> > **Note:** Small negative values can occur for wSMI, reflecting estimator noise or a lack of information sharing above chance, rather than an anti-coupling of signals.
> 
> What do you think?

Great, I think that would help

Laouen:
> > Hi @tsbinns sorry for not being able to connect in your last office hours. Sadly @GioMarraffini is currently on a conference so we can't yet have a time.
> > Can you give us the calendar with the next office hours so we organize ourself to make a slot and finally do that call ?
> 
> Hi @Laouen, the next office hours isn't until the 19th (calendar link here: https://mne.discourse.group/t/mne-office-hours-on-discord/7439). I finally have time this week free, so I will work on the tests and get things into a more mergeable state.

@tsbinns Perfect, thanks for the calendar link. @GioMarraffini and I are going to connect in the next office hours (12/19) to the discor so we can have the meeting with you. Looking forward to discuss this PR with you!

examples/wsmi.py
```diff
@@ -0,0 +1,457 @@
...
+#
+# **``anti_aliasing=False``:**
+#
+# - Never applies filtering - use only if you've already filtered appropriately
+# - **Warning**: Results may be inaccurate without proper filtering
+#
+# The tau parameter affects your effective sampling frequency multiplicatively. For
+# example, ``tau = 2`` reduces your effective sampling rate by half, so
+# high-temporal-fidelity coupling (sample-to-sample) may not be detected if you set tau
+# to higher values. This affects the time-scale on which signals are considered coupled.
```
tsbinns: Re-reading the anti-aliasing section, this note on tau and effective sampling frequency seems a little disjointed. Can we make the link between to anti-aliasing a bit more concrete?

tsbinns (changes requested):
Thanks for the changes @GioMarraffini & @Laouen! The anti-aliasing and weak/strong tests are great. I also just pushed a few small changes myself.

My only suggestions are two things to clarify in the docs, and a request for a test that the auto anti-aliasing mode works as expected.
Also, please add a little note about the origin of the testing data as per this comment: https://github.com/mne-tools/mne-connectivity/pull/307#discussion_r2635732124

examples/wsmi.py
```diff
@@ -0,0 +1,460 @@
...
+                f"{data[ii, jj]:.3f}",
+                ha="center",
+                va="center",
+                color="black" if data[ii, jj] > data.max() / 2 else "white",
+                fontsize=8,
+            )
+    plt.colorbar(im, ax=axes[idx], label="wSMI")
+
+plt.suptitle("Parameter Grid Search: wSMI Recovery for Different Coupling Types")
+plt.tight_layout()
```
tsbinns: Re: https://github.com/mne-tools/mne-connectivity/pull/307#discussion_r2294278657

Source <-> Independent connectivity looked as high as Source <-> Coupled because connectivity was not being indexed correctly. Have fixed now.

tsbinns (approved):
@GioMarraffini @Laouen Have had a chance to look over the anti-alisaing test and I'm happy with it. With that, I'm going to approve. Thanks so much for the great contribution and sticking with this over the last months!

@drammock I've given everything the **twice** over. Given how much we both poured over this before, I think we're good to go, but I won't (dismiss your review and) merge right away in case you want to double check anything.

@Laouen, I added your name to the changelog entry too, and your name with GH tag to the authors file. Push if you want this changed before merging!

tsbinns:
Docs failure due to osf timeout

PR #326, opened 2025-09-04
[API] Change arbitrary `sfreq` default value in `phase_slope_index`

### LINKED ISSUE #324: Arbitrary `sfreq` default value used in `phase_slope_index`
`phase_slope_index` uses the arbitrary default `sfreq=2*np.pi`.

This was raised way back in #9 and was updated for `spec_conn_epochs` in #119 to use `None` as default, requiring `sfreq` to be provided with array data. However, the same change was never made for `phase_slope_index`.

What are peoples' thoughts on updating the API to use `sfreq=None` by default? I am happy to open a PR with a deprecation warning.

tsbinns:
PR Description
--------------

Fixes #324

Changes `phase_slop_index` docs to state that `sfreq=None` is supported and that this will replace the current default `2*np.pi` in v0.9.

Adds a corresponding `FutureWarning` for when the current default `sfreq` is used with array-like data (parameter is ignored for `Epochs`).

### DIFF
--- a/mne_connectivity/effective.py
+++ b/mne_connectivity/effective.py
@@ -5,7 +5,7 @@
 import copy
 
 import numpy as np
-from mne.utils import logger, verbose
+from mne.utils import logger, verbose, warn
 
 from .base import SpectralConnectivity, SpectroTemporalConnectivity
 from .spectral import spectral_connectivity_epochs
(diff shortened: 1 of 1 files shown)


mne_connectivity/effective.py
```diff
@@ -120,6 +122,15 @@ def phase_slope_index(
     .. footbibliography::
     """
     logger.info("Estimating phase slope index (PSI)")
+
+    if sfreq == 2 * np.pi and isinstance(data, np.ndarray | list | tuple | set):
```
larsoner, line 126: This isn't quite right I think. If a user themselves passed 2*pi correctly, they'll get an incorrect warning.

Instead you can change the default to, say, an empty string. Detect if it's an empty string here and enjoy a warning and set to 2*pi.

tsbinns: Fair point, have just pushed a change for this.

mne_connectivity/tests/test_connectivity.py
```diff
@@ -392,7 +392,11 @@ def test_metadata_handling(func, tmpdir, epochs):
     an ``mne.Epochs`` object input.
     """
     kwargs = dict()
-    if isinstance(epochs, np.ndarray) and func == spectral_connectivity_epochs:
+    if (
+        isinstance(epochs, np.ndarray)
+        and func == spectral_connectivity_epochs
```
larsoner, line 397: `and` is higher precedence than `or` so I think you want parentheses here? Or use `or func in (x, y)`

larsoner:
Thanks @tsbinns !

PR #334, opened 2025-09-17
[BUG] Fix including bad channels in computations

### LINKED ISSUE #323: Connectivity methods including bad channels in computations
#### Describe the bug
Noted here: https://github.com/mne-tools/mne-connectivity/pull/307#discussion_r2294725022

When `indices=None` or when no `indices` parameter is supported, connectivity is computed using all available channels. It can be expected that bad channels would not be included to avoid unnecessary computations.

However, in `spectral_connectivity_epochs`, `spectral_connectivity_time`, `phase_slope_index`, `vector_auto_regression`, and `envelope_correlation`, bad channels are considered as valid and included in connectivity computations.

#### Possible solution
When `indices=None` or when no `indices` parameter is supported (e.g., `vector_auto_regression`), ignore bads when calling `Epochs.get_data()` and make sure other information for bad channels extracted from the `Epochs` objects is dropped (e.g., from channel names).

Bads are dropped by default when calling `get_data` from `EpochsSpectrum/TFR` objects, so any existing logic to force the inclusion of bad channels for consistency with the previous `Epochs.get_data()` behaviour would need to be adjusted accordingly.

Based on the original comment, this could be considered unexpected behaviour, so would I be correct in saying we can change this without a deprecation cycle?

tsbinns:
PR Description
--------------

Fixes #323.

When `indices=None` (`spec_conn_epochs/time`, `phase_slope_index`) or when no `indices`  parameter is supported (`envelope_correlation`, `vector_auto_regression`), bad channels are now ignored from connectivity computations.

However, entries for these bad channels will still be considered in the dense representation of the matrices as per this discussion: https://github.com/mne-tools/mne-connectivity/issues/323#issuecomment-3286402064

### DIFF
--- a/mne_connectivity/envelope.py
+++ b/mne_connectivity/envelope.py
@@ -9,6 +9,7 @@ import inspect
 
 import numpy as np
 from mne import BaseEpochs
+from mne._fiff.pick import _picks_to_idx
 from mne.filter import next_fast_len
 from mne.source_estimate import _BaseSourceEstimate
 from mne.utils import _check_option, _ensure_int, _validate_type, logger, verbose, warn
@@ -75,11 +76,15 @@ def envelope_correlation(
 
     corrs = list()
 
+    picks = None
     n_nodes = None
 
     events = None
     event_id = None
     if isinstance(data, BaseEpochs):
+        # Find good channels
+        picks = _picks_to_idx(data.info, picks="all", exclude="bads")
+
         names = data.ch_names
         events = data.events
         event_id = data.event_id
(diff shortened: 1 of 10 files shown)


mne_connectivity/envelope.py
```diff
@@ -75,11 +76,15 @@ def envelope_correlation(
...
 
+    picks = None
     n_nodes = None
 
     events = None
     event_id = None
     if isinstance(data, BaseEpochs):
+        # Find good channels
+        picks = _picks_to_idx(data.info, picks="all", exclude="bads")
+
```
tsbinns, line 87: I am using this to identify good channels, but I'm wondering whether `picks="all"` is the best choice here (also `with_ref_meg=True` by default).
E.g., I saw in some PSD computations in MNE-Python that `picks="data_or_ica"` and `with_ref_meg=False` were used.

larsoner: I think it's okay to use all here -- users should pass `picks` themselves (and we should make that an option) or do `epochs.pick` themselves ahead of time

larsoner (approved):
Thanks for the explanations, all looks good to me @tsbinns !

PR #338, opened 2025-10-01
[MAINT] Update path for CircleCI artifacts redirector action

tsbinns:
Discussed in meeting with @larsoner.


larsoner:
Only takes effect once merged into `main` so let's do it!

PR #405, opened 2026-04-08
Add missing PR credit

tsbinns:
A number of contributors are present in the changelogs but missing credit in git for these PRs, causing these contributors to be missing from `CITATION.cff`. Fixes this by adding fake PRs for those authors.

@drammock, this was everything I noticed.


tsbinns:
@drammock Realising this also pumps my contrib numbers even though I didn't specify myself as an author.

drammock:
> @drammock Realising this also pumps my contrib numbers even though I didn't specify myself as an author.

yeah I knew it would do that, but you're already the front-runner and nobody is likely to catch up anytime soon, so it probably won't ever matter (in terms of contributor order in the CFF file)

PR #403, opened 2026-04-08
List `cohy` as a supported method in `spectral_connectivity_time`

tsbinns:
Fixes #389.

Mainly just adding missing documentation. But also adds tests for `cohy` in `spectral_connectivity_time` to be thorough.

Would be a good backporting candidate for a v0.8.1.

PR #410, opened 2026-04-11
Fix changelog

tsbinns:
Removed the changelog entry for #403 which was backported to v0.8.1.

Also prevents dev changelog title link appearing as "Version Version X.Y.Z...".

### DIFF
--- a/doc/changes/dev/403.other.rst
+++ b/doc/changes/dev/403.other.rst
@@ -1,1 +1,0 @@
-List ``cohy`` as a supported method in :func:`~mne_connectivity.spectral_connectivity_time`, by `Thomas Binns`_.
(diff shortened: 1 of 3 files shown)


tsbinns:
Not sure why changelog check is failing: https://github.com/mne-tools/mne-connectivity/actions/runs/24281579938/job/70904197707?pr=410#step:4:15

Looking at the code for the action, is it because a changelog entry has been modified but the `no-changelog` label is present?
https://github.com/scientific-python/action-towncrier-changelog/blob/798485ae347fb4798b3fcbe9e4f7f290a2d2344b/check_changelog.py#L234-L236

tsbinns:
Need to re-add the dropped #403 changelog entry to the v0.8 changes

larsoner:
Yes what you probably want is the skip-changelog-checks label

PR #411, opened 2026-04-11
Update contributing guide

tsbinns:
Fixes #386 by providing info on online docs building & deployment.

Also updates release info for dynamic versioning and micro releases.


larsoner:
Thanks @tsbinns !

FYI in MNE-Python we keep the release notes in the Wiki and just tweak steps as needed. But fine to have it in the CONTRIBUTING.md here

PR #412, opened 2026-04-11
Fix imag data being discarded for `get_data` with dense output

tsbinns:
Found when playing around with stuff for #372.

When `output="dense"` used with `get_data`, the data gets returned as an array with a new shape.

However, this wasn't taking into account data's dtype, so complex-valued data was getting stuffed into a real-valued array.

Adds some tests with this that would fail on main.


larsoner (approved):
Good catch, thanks @tsbinns !

PR #415, opened 2026-04-26
Add surrogate generation from evoked data

### LINKED ISSUE #251: Within-epoch surrogate generation
#### Describe the problem

It is difficult to interpret many connectivity estimates without comparison to "null" surrogate data. As stated in [Aru et al., 2015](https://www.sciencedirect.com/science/article/pii/S0959438814001640#bib0015) : "A suitable surrogate construction should only destroy the specific cyclo-stationarities related to the hypothesized... effect, while keeping all the unspecific non-stationarities and non-linearities of the original data."

#### Describe your solution
It would make sense to perform trial/epoch level scrambling so that non-stationarities and phase distributions within a trial/epoch are conserved. [Tensorpac](https://github.com/EtienneCmb/tensorpac/blob/ac9058fd375d423757943810f613d63785fab85f/tensorpac/methods/meth_surrogates.py#L59) achieves this with a simple function, which I have edited slightly:

```
def swap_time_blocks(data, random_state=None):

    """Compute surrogates by swapping time blocks.
    This function cuts the timeseries at a random time point. Then, both time
    blocks are swapped.
    Parameters
    ----------
    data : array_like
        Array of shape (n_chan, ..., n_times).
    random_state : int | None
        Fix the random state of the machine for reproducible results.
    Returns
    -------
    surr : array_like
        Swapped timeseries to use to compute the distribution of
        permutations
    References
    ----------
    Source: https://www.sciencedirect.com/science/article/pii/S0959438814001640
    """
    
    if random_state is None:
        random_state = int(np.random.randint(0, 10000, size=1))
    rnd = np.random.RandomState(random_state)
    
    # get the minimum / maximum shift
    min_shift, max_shift = 1, None
    if not isinstance(max_shift, (int, float)):
        max_shift = data.shape[-1]
    # random cutting point along time axis
    cut_at = rnd.randint(min_shift, max_shift, (1,))
    # split amplitude across time into two parts
    surr = np.array_split(data, cut_at, axis=-1)
    # revered elements
    surr.reverse()
    
    return np.concatenate(surr, axis=-1)
```

Example usage with mne data: 

```
    data = np.swapaxes(mne_data.get_data(copy=False), 0, 1) # swap so now it's chan, events, times 
    surr_dat = np.zeros_like(data) # allocate space for the surrogate channels 
    for ix, ch_dat in enumerate(data): # apply the same swap to every event in a channel, but differ between channels 
        surr_ch = swap_time_blocks(ch_dat, random_state=None)
        surr_dat[ix, :, :] = surr_ch
    surr_dat = np.swapaxes(surr_dat, 0, 1) # swap back so it's events, chan, times 
    # make a new EpochArray from it
    surr_mne = mne.EpochsArray(surr_dat, 
                mne_data.info, 
                tmin=mne_data.tmin, 
                events = mne_data.events, 
                event_id = mne_data.event_id)
```

tsbinns:
Fixes #251

- Adds `make_surrogate_evoked_data` for generating surrogates from evoked data.
- Updates the example to demonstrate the new function.
- Adds `make_surrogate_resting_data` which will replace `make_surrogate_data` in a later release.
- Adds support for `Epochs` objects to `make_surrogate_resting_data`.

Main thing missing is a test that validates the `make_surrogate_evoked_data` approach. Our existing data simulation only generates pseudo-resting data, but we need to be testing on evoked data.

We could use something from an example dataset (e.g., somato - like we do for the example). We could also look at adding functionality for simulating evoked data, but that's a much more considerable undertaking and I'm not sure how much use there would be for it outside of this.

Will leave in draft until the testing situation is clearer.


tsbinns:
Any diff related to evoked data sim should disapper once #418 is merged.

tsbinns:
Will keep in draft until #418 merged

tsbinns:
Screwed up the rebase so commit log is a mess, but diff is now aligned with #418 so this will be ready for review once that's merged.

tsbinns:
@drammock Diff is simpler now that #418 is merged.

One thing I noticed was a few references to 0.10 being the next version, but as I understand this would now be 1.0. Docs build is also failing due to RAM limits on the surrogate connectivity example.

I will sort these things out before asking for your review.

tsbinns:
Profiled memory usage and max memory consumption is 1 GB. Seems actually `n_jobs=None` in `compute_tfr()` meant joblib was being used and the resources weren't cleaned up properly. Setting `n_jobs=1` fixes this.

tsbinns:
@drammock Made some final tweaks. This is ready for review now.

Test coverage for new features is there. The diff for the testing goes a bit beyond just checking the new features, but also cleaning things up and making the tests more robust in general, which I mentioned before.

examples/surrogate_connectivity.py
```diff
@@ -77,74 +81,80 @@
...
-# the baseline connectivity, we perform this shuffling procedure
-# :math:`\text{n}_{\text{shuffle}}` times, producing :math:`\text{n}_{\text{shuffle}}`
-# surrogate datasets. We can then iterate over these shuffles and compute the
-# connectivity for each one.
+# Next, we generate the surrogate data by passing the coefficients into the
+# :func:`~mne_connectivity.make_surrogate_resting_data` function. This approach
+# destroys the data's covariance structure by randomly shuffling the order of epochs,
+# independently for each channel. To get a reliable estimate of the baseline
+# connectivity, we perform this shuffling procedure :math:`\text{n}_{\text{shuffle}}`
+# times, producing :math:`\text{n}_{\text{shuffle}}` surrogate datasets. We can then
```
drammock: why not `n_shuffles` (formatted as code, like the param name) instead of ``:math:`\text{n}_\text{shuffle}` `` here?

examples/surrogate_connectivity.py
drammock: there's a missing closing bracket on this line
```suggestion
# \text{ if otherwise } \}` ,
```

examples/surrogate_connectivity.py
```diff
@@ -185,16 +195,14 @@
...
-beta_con_surrogate = np.abs(
-    [surrogate.get_data()[:, beta_freqs] for surrogate in surrogate_con]
-).mean(axis=(1, 2))
+# Compute lower beta connectivity for surrogate data
+beta_con_pretrial_surrogate = pretrial_surrogate_con[:, beta_freqs].mean(axis=1)
 
 # Compute p-value for pre-trial lower beta coupling
-p_val = np.sum(beta_con_pretrial <= beta_con_surrogate) / n_shuffles
+p_val = np.sum(beta_con_pretrial <= beta_con_pretrial_surrogate) / n_shuffles
 print(f"P = {p_val:.2f}")
```
drammock: rounding to 2 digits is an odd choice as it yields `p=0.00`.  Can you format in scientific notation rounded to 2 digits instead? e.g. `p=1.2e-4` or whatever?

(also: p should be lowercase)

same comments for the evoked surrogate section below

examples/surrogate_connectivity.py
```diff
@@ -204,120 +212,152 @@
...
 ########################################################################################
 # .. _inappropriate-surrogate-data:
 #
 # Generating surrogate connectivity from inappropriate data
 # ---------------------------------------------------------
-# We discussed above how surrogates generated from evoked data risk overestimating the
-# degree of baseline connectivity. We demonstrate this below by generating surrogates
-# from the post-stimulus data.
+# We discussed above how using evoked data to generate surrogates from
+# :func:`~mne_connectivity.make_surrogate_resting_data` risks overestimating the degree
```
drammock:
```suggestion
# We discussed above how using :func:`~mne_connectivity.make_surrogate_resting_data`
# to estimate a null connectivity distribution for *evoked data*
# risks overestimating the degree
```

examples/surrogate_connectivity.py
drammock:
```suggestion
# single-recording-level, we may also want to determine whether group-level connectivity
```
(just a missing hyphen)

mne_connectivity/datasets/surrogate.py
```diff
@@ -82,80 +151,236 @@ def make_surrogate_data(data, n_shuffles=1000, rng_seed=None, return_generator=T
...
+    ----------
+    data : ~mne.Epochs | ~mne.time_frequency.EpochsTFR
+        The time-series or Fourier coefficients to create the null hypothesis surrogate
+        data for. Fourier coefficients can be generated from
+        :meth:`mne.Epochs.compute_tfr` with ``output='complex'``.
+    n_shuffles : int (default 1000)
+        The number of surrogate datasets to create.
+    rng_seed : int | None (default None)
+        The seed to use for the random number generator. If ``None``, no seed is
+        specified.
```
drammock, line 204: I think in MNE-Python, we allow users to pass anything that the underlying `np.default_rng(seed=)` will accept... which I think means also tuple of ints, `np.random.Generator` instance, and maybe some other stuff.  Not critical to change in this PR, but let's put it on the 1.0 roadmap to make sure we're being consistent about this across MNE-Connectivity and MNE-Python

drammock (approved):
all minor comments; +1 to merge when addressed/rebutted.

examples/surrogate_connectivity.py
tsbinns: My intention was more to have something like this:
image

I can try to format that properly.

examples/surrogate_connectivity.py
tsbinns: For this toy example with only 100 shuffles, there are no cases where the surrogate is > real data in that freq band, so p-val is just 0.

mne_connectivity/datasets/surrogate.py
tsbinns: Good point! Have opened #435 as a reminder on the 1.0 milestone.

examples/surrogate_connectivity.py
tsbinns: Looks like this now:
image

tsbinns:
@drammock Have addressed/rebutted those comments. Please LMK if you're happy with state of things now.

drammock:
Looks good thanks @tsbinns

PR #418, opened 2026-05-06
Add ability to simulate connectivity in a specific time range

tsbinns:
Preparation to simplify diff in #415. @drammock

Adds ability to simulate connectivity in specific time ranges in `make_signals_in_freq_bands`.

Also makes some improvements to the unit test. Before, we were testing two SNRs, but not doing any comparison that connectivity was greater with the higher SNR. Now, we're checking that the higher SNR can be better distinguished from the noise. It also means we can simplify the expected connectivity values.


mne_connectivity/datasets/frequency.py
```diff
@@ -51,6 +56,25 @@ def make_signals_in_freq_bands(
         Number of timepoints for the delay of connectivity between the seeds and
         targets. If > 0, the target data is a delayed form of the seed data. If < 0, the
         seed data is a delayed form of the target data.
+    connection_time : float | None (default None)
+        Time at which the interaction occurs in each epoch, in seconds. Must be greater
+        than ``tmin`` and less than ``n_times / sfreq``. If ``None``, the connectivity
+        is simulated throughout the whole epoch.
```
tsbinns: My reservation is that I'm working here in seconds, but for other time-related params in samples (e.g., `connection_delay`).
I think seconds is the most intuitive, and I'm not sure why we intially settled on using samples, but it does introduce a mix of time units.

tsbinns:
Not sure if we should demo this functionality by modifying an example.

larsoner:
> Not sure if we should demo this functionality by modifying an example.

Would be cool... maybe as a way to show some differences between a few connectivity measures?

> My reservation is that I'm working here in seconds, but for other time-related params in samples (e.g., connection_delay).

Agree seconds is most natural. Can other params be converted via deprecation? Or would this be too annoying?

tsbinns:
> Would be cool... maybe as a way to show some differences between a few connectivity measures?

Okay, will have a think about where this could sit.

> Agree seconds is most natural. Can other params be converted via deprecation? Or would this be too annoying?

If `connection_delay` is changed alone, I'm not sure how we could allow users to prepare for the new seconds behaviour whilst not immediately breaking the existing samples behaviour.
However, if `n_times` were also deprecated in favour of something like `duration` for specifying epoch lengths: if `n_times` is used, we could have `connection_delay` recognised in units of samples; if `duration` is used, `connection_delay` is recognised in units of seconds. That way we can maintain existing behaviour for the deprecation cycle, but allow users to switch to the new behaviour.

tsbinns:
> Would be cool... maybe as a way to show some differences between a few connectivity measures?

Didn't find a great place this could sit and didn't really want to make a whole new example (esp. when the new functionality is pretty straight forward).

Have instead added a small code block to the notes section of the docstring showing how this can be used.

mne_connectivity/datasets/frequency.py
tsbinns: Have begun depr in favour of seconds

tsbinns:
@drammock The deprecation from samples to seconds is now added, so this is ready for review. The deprecation adds a bit of noise to other examples/tests, but I've just switched it to be the equivalent behaviour using seconds as it was using samples.

There are also some big changes to the simulation unit tests. If you recall I mentioned before that this just simplifies some logic and makes the tests more valid.

All new features have test coverage.

Once this is merged, #415 would be ready for review with a simpler diff (although merging will probably require resolving some conflicts).

mne_connectivity/datasets/frequency.py
```diff
@@ -120,21 +238,39 @@ def make_signals_in_freq_bands(
         fir_design="firwin2",
     )
 
+    # isolate interaction to a given time range per epoch
+    if connection_time is not None:
+        connection_time -= tmin  # convert time to be relative to time of first sample
+        time_samples = (connection_time * sfreq).astype(int)  # convert time to samples
```
drammock: `.astype(int)` risks converting e.g. 249.999 into 249.  Better to round first and then cast.

mne_connectivity/decoding/tests/test_decomposition.py
```diff
@@ -49,10 +51,11 @@ def test_spectral_decomposition(method, mode):
         n_seeds=n_seeds,
         n_targets=n_targets,
         freq_band=(fmin_ignore, fmax_ignore),
+        duration=duration,
         n_epochs=n_epochs,
         trans_bandwidth=trans_bandwidth,
         snr=0.5,
-        connection_delay=6,  # ~90° interaction angle for this freq. band
+        connection_delay=0.06,  # ~90° interaction angle for this freq. band
```
drammock: was slightly confused that both tests have this same comment but different delays.  Quickly realized it's because the bands are different (and thus the mean period different) but might avoid future maintainer confusion to say "~90° interaction angle for a 21-24Hz band" here (and similarly for the prior comment)

drammock (approved):
LGTM, just 2 small nitpicks/comments. Thanks @tsbinns and sorry for the delay in reviewing this!

tsbinns:
@drammock Have adressed your comments. Will merge #432 to fix doc building, then this should be able to go in.

drammock (approved):

drammock:
great, +1 for merge when green

PR #420, opened 2026-05-17
Fix `DynamicMixIn` for VAR models

### LINKED ISSUE #332: Cannot predict from VAR model with lags > 1
#### Describe the bug
Found while working on #330.

The `predict()` method of connectivity objects cannot be used when VAR coeffs have multiple lags.
I haven't been able to investigate exactly why this is occurring or if this behaviour was ever supported, but I wanted to open this as a reminder to myself.

#### Steps to reproduce

```
import numpy as np

from mne_connectivity import vector_auto_regression

n_epochs, n_signals, n_times = 2, 3, 64
data = rng.randn(n_epochs, n_signals, n_times)
times = np.arange(n_times)

conn = vector_auto_regression(data, model="dynamic", lags=2)
conn.predict(data)
```

#### Expected results

The predicted data is returned.

#### Actual results

```
File "\GitHub\mne-connectivity\mne_connectivity\base.py", line 282, in predict      
    predicted_data[jdx, :, lags:] += np.dot(
                                     ^^^^^^^
ValueError: shapes (3,2,2) and (3,62) not aligned: 2 (dim 2) != 3 (dim 0)
```

tsbinns:
Fixes #332.

Right now corrects the indexing issue, extends some of the data validation, and tightens the objects supporting the `predict` method (shouldn't be available for spectral data).

The unit tests will need to be adjusted further, mainly with some new validations that the predictions return expected results. One idea for that is to compute Granger causality, with the residuals derived from these predictions. A simpler option would be just looking that the residuals we get from a `dynamic` model is smaller than from an `avg-epochs` model.


mne_connectivity/base.py
```diff
@@ -191,126 +201,168 @@ def eigvals(self):
...
-            This function is used to create the generating noise process. If ``None``,
+            The function used to create the generating noise process. Each call to this
+            function should return an array of shape ``(n_nodes,)``. If ``None``,
             Gaussian white noise with zero mean and unit variance is used.
         %(random_state)s
 
         Returns
         -------
-        data : array, shape (n_samples, n_channels)
+        data : array, shape (n_nodes, n_samples)
```
tsbinns, line 342: Just a doc fix, the behaviour hasn't changed

mne_connectivity/base.py
```diff
@@ -191,126 +201,168 @@ def eigvals(self):
...
             Generated data.
+
+        Notes
+        -----
+        If the VAR model is time-varying (i.e., there are different models per epoch),
+        the data is simulated from the average of these models.
```
tsbinns, line 348: Just a doc clarification of the existing behaviour

mne_connectivity/base.py
```diff
@@ -191,126 +201,168 @@ def eigvals(self):
...
-                    )
+        for lag_idx in range(1, lags + 1):
+            for epo_idx in range(n_epochs):
+                var_idx = epo_idx if self.is_epoched else 0
+                bp = var_model[var_idx, :, (lag_idx - 1) :: lags]
+                predicted_data[epo_idx, :, lags:] += np.dot(
+                    bp, data[epo_idx, :, (lags - lag_idx) : (n_times - lag_idx)]
+                )
+        if data_ndim == 2:  # remove unwanted epochs dim
+            predicted_data = predicted_data[0]
```
tsbinns, line 322: Is a behaviour change, but changes it so that it aligns with what the docs state

mne_connectivity/base.py
```diff
@@ -191,126 +201,168 @@ def eigvals(self):
...
-                    predicted_data[jdx, :, lags:] += np.dot(
-                        bp, data[jdx, :, (lags - idx) : (n_times - idx)]
-                    )
+        for lag_idx in range(1, lags + 1):
+            for epo_idx in range(n_epochs):
+                var_idx = epo_idx if self.is_epoched else 0
+                bp = var_model[var_idx, :, (lag_idx - 1) :: lags]
+                predicted_data[epo_idx, :, lags:] += np.dot(
+                    bp, data[epo_idx, :, (lags - lag_idx) : (n_times - lag_idx)]
+                )
```
tsbinns, line 320: Switches to using exclusively the 2nd option since: 1) in my experience this is the more common scenario (`n_epochs < n_times - lags`); 2) more easily allows for new feature of predictions to be made from epoched data when the VAR model is not time-varying; and 3) some indexing within the 1st option doesn't work anyway.

drammock (approved):

PR #430, opened 2026-06-21
Fix slow pytest 9.1.x

### LINKED ISSUE #429: Figure out why Pytest 9.1.x is slow
> Yeah, reverting to pytest 9.0.3 fixes it. Will add a req to avoid 9.1.0 for now. Don't see any mention of problems like this in their gh issues. Latest test runs for MNE-Python I saw weren't using 9.1.0 yet, so will be interesting to see if it slows there as well.

_Originally posted by @tsbinns in https://github.com/mne-tools/mne-connectivity/issues/427#issuecomment-4717732007_

tsbinns:
Fixes #429

Locally, test suite runs with pytest 9.0.3 and 9.1.1 take the same time to run (Ubuntu 24.04, Python 3.13).

However, with pytest 9.1.1, `test_plot_connectivity_circle` (the last test that gets run) has a plot pop-up which I have to close before the tests can finish, and which I didn't see for 9.0.3.

There is a fixture that is supposed to prevent this from happening, but maybe something in 9.1.1 is preventing the fixture from working properly.

This is the only visualisation test that actually gets run, as the others are skipped due to lack of imageio-ffmpeg

Have disabled the test in question to see if that's indeed the culprit.

### DIFF
--- a/mne_connectivity/conftest.py
+++ b/mne_connectivity/conftest.py
@@ -85,7 +85,7 @@ def close_all():
     plt.close("all")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="session", autouse=True)
 def matplotlib_config():
     """Configure matplotlib for viz tests."""
     import matplotlib
--- a/mne_connectivity/viz/tests/test_circle.py
+++ b/mne_connectivity/viz/tests/test_circle.py
@@ -13,6 +13,7 @@ from mne.viz import circular_layout
 from mne_connectivity.viz import plot_connectivity_circle
 
 
+@pytest.mark.skip
 def test_plot_connectivity_circle():
     """Test plotting connectivity circle."""
     node_order = [
--- a/mne_connectivity/viz/tests/test_circle.py
+++ b/mne_connectivity/viz/tests/test_circle.py
@@ -5,7 +5,6 @@
 # License: Simplified BSD
 
 
-import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 from mne.viz import circular_layout
@@ -13,7 +12,6 @@ from mne.viz import circular_layout
 from mne_connectivity.viz import plot_connectivity_circle
 
 
-@pytest.mark.skip
 def test_plot_connectivity_circle():
     """Test plotting connectivity circle."""
     node_order = [
@@ -176,4 +174,3 @@ def test_plot_connectivity_circle():
     pytest.raises(
         ValueError, circular_layout, label_names, node_order, group_boundaries=[20, 0]
     )
-    plt.close("all")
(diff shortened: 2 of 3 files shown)


tsbinns:
Yeah, tests run in normal time with that disabled, so tests aren't taking longer, it just gets caught on the final one.

tsbinns:
The logic in this `conftest.py` matches what is in MNE, but it didn't seem to be taking effect.

Explicitly adding `autouse=True` in the matplotlib config fixture prevented the figure popping up locally, but this should not be necessary because it gets added as a used fixture to the config (as is done for MNE).

However, it also resolved locally when I explicitly added `mne_connectivity` to the `testpaths` config option.

I'm not sure which change exactly in 9.1.0 caused the fixture to no longer take effect, but it runs now 🤷🏻

tsbinns:
Removed the pytest specs to avoid 9.1. Most of the diff is the formatter changing pyproject.toml

tsbinns:
Huh, updating testpaths fixed it locally, but not here. Seeing if adding autouse to the fixture also only works locally.

tsbinns:
Setting the fixture to autouse fixes it on runners and locally. However, I would like to know why the config changes we make in conftest don't seem to take effect, so I'll wait before merging this.

mne_connectivity/viz/tests/test_circle.py
```diff
@@ -5,7 +5,6 @@
 # License: Simplified BSD
 
 
-import matplotlib.pyplot as plt
```
larsoner, line 8: It's possible that doing this at the top sets the backend to non-Agg, which could be slower. I'd be surprised if it's 10x slower but 🤷 

In MNE-Python we force set to Agg I think unless an env var is set to use Qt (which can be useful for debugging). I'd force-set here as well if that's not done already (haven't looked!)

mne_connectivity/conftest.py
larsoner, line 105: Oh nevermind this is already force-setting to `agg`

larsoner:
> Setting the fixture to autouse fixes it on runners and locally. However, I would like to know why the config changes we make in conftest don't seem to take effect, so I'll wait before merging this.

My shot in the dark is that in previous versions, having a session-scoped fixture might have caused it to be used even if it's not marked `autouse`. So arguably it's a bugfix that you now have to specify `autouse=True` even if it's not used in any tests

larsoner:
> but this should not be necessary because it gets added as a used fixture to the config (as is done for MNE).

Okay I didn't read everything until just now... yeah this part seems weird. You could open a bug report about this change in behavior if you can isolate it properly.

Can you make the corresponding `autouse` change in MNE-Python as well? I don't see a problem with adding it both places. Then you could remove it from where it's told to be used in the `config`

PR #432, opened 2026-06-24
Bump CircleCI runner image

tsbinns:
Docs can no longer build after https://github.com/mne-tools/mne-python/pull/13978 (changed `/tools/circleci_bash_env.sh`, which we use).

Try to adapt the changes for MNE-Conn here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   build_docs:
     machine:
-      image: ubuntu-2404:current
+      image: ubuntu-2604:2026.05.1
     steps:
       - restore_cache:
           keys:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,8 +19,6 @@ jobs:
       - uses: actions/checkout@v7
       - name: Set up Python
         uses: actions/setup-python@v6
-        with:
-          python-version: '3.10'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip


.github/workflows/release.yml
```diff
@@ -19,8 +19,6 @@ jobs:
       - uses: actions/checkout@v7
       - name: Set up Python
         uses: actions/setup-python@v6
-        with:
-          python-version: '3.10'
```
tsbinns, line 23: Also copies the change of using the runner default Python ver which gets auto-bumped. This is the only case I found.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 jobs:
   build_docs:
     machine:
-      image: ubuntu-2404:current
+      image: ubuntu-2604:2026.05.1
```
larsoner, line 6: Argh in principle you shouldn't need this change, it means we should update MNE-Python. I tried to make it 24.04 compatible with a conditional but I must have messed it up 🤦 (And really we should move this to the shared repo where we can test it runs on 24.04 and 26.04 for example!)

I am getting a similar failure in mne-gui-addons...

Do you want to try to fix the MNE-Python script? If not I can do it.

larsoner: ... the error in `mne-gui-addons` is:
```
      File "/home/circleci/python_env/lib/python3.10/site-packages/mne_gui_addons/_vol_stc.py", line 361, in __init__
        ) = self._renderer._volume(
    TypeError: _PyVistaRenderer._volume() missing 1 required positional argument: 'interpolation'
```
which is probably separate actually but can be fixed by setting the default there to `interpolation="linear"` (plus a note saying that this private method is used by `mne-gui-addons` so we have to be careful about backward incompat changes...)

tsbinns: The error I was getting was `Unable to locate package libgvplugin-neato-layout8` from the `circleci_bash_env.sh` call, which was a newly added install.

So, would it be a similar change to what you did for `setup_xvfb.sh`, where we just have different installs for 24.04 and 26.04?
https://github.com/larsoner/mne-python/blob/9c50ccbda3f82c233ad14e500e9705e161348799/tools/setup_xvfb.sh

larsoner: Yeah that should do it! ... Plus a `_volume(..., interpolation="linear")` fix while you're at it if you're up for it :pray:

tsbinns: Sure, can do now!

tsbinns: Opened https://github.com/mne-tools/mne-python/pull/13985

tsbinns:
Switching to 26.04 no longer required after https://github.com/mne-tools/mne-python/pull/13985, but I don't think there's harm in bumping.

PR #433, opened 2026-06-24
Clean up CircleCI run

tsbinns:
Installation check was duplicated.

Some example data was also not being installed up front.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,9 +68,17 @@ jobs:
       - restore_cache:
           keys:
             - data-cache-sample
+      - restore_cache:
+          keys:
             - data-cache-bst-resting
+      - restore_cache:
+          keys:
             - data-cache-somato
+      - restore_cache:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -106,6 +96,9 @@ jobs:
             mkdir -p ~/mne_data
             python -c "import mne; print(mne.datasets.sample.data_path(update_path=True))";
             python -c "import mne; print(mne.datasets.brainstorm.bst_resting.data_path(update_path=True))";
+            python -c "import mne; print(mne.datasets.somato.data_path(update_path=True))";
+            python -c "import mne; print(mne.datasets.fieldtrip_cmc.age.data_path(update_path=True))";
+            python -c "import mne; print(mne.datasets.epilepsy_ecog.data_path(update_path=True))";
```
larsoner, line 108: you should probably add matching cache entries for these, too

PR #436, opened 2026-07-13
Expand main content area

tsbinns:
Copying changes from https://github.com/mne-tools/mne-python/pull/14015 here.

Having to hscroll for some of the code cells is something that had annoyed me before.

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -73,15 +73,3 @@ p.logo__title {
 .bd-page-width {
     max-width: 100rem;
 }
-
-.bd-sidebar-primary {
-    max-width: 16rem;
-}
-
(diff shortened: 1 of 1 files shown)


tsbinns:
Formatter added a lot of whitespace, but last block is the only actual addition.

drammock:
Feel free to self merge if it looks how you expect, I won't have time to look this week

tsbinns:
Max width of 16 rem for the primary sidebar is too compressed IMO. I think the default 25% looks better (and then the default 60 rem for the article container max width also works well). Also think having non-justified main content is better, since when there is no primary sidebar, the content is too far on the left.

So only change is to take max page width from 88 to 100 rem.

PR #439, opened 2026-07-21
Propose adding Kalofolias (2016) graph-learning connectivity method

payamsash:
PR Description
--------------

This PR proposes adding a graph-learning-based connectivity estimator based on this [paper](https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=Vassilis+Kalofolias.+How+to+learn+a+graph+from+smooth+signals.+In+Artificial+intelligence+and+statistics%2C+920–929.+PMLR%2C+2016.&btnG=).

The method estimates a sparse weighted graph directly from signal smoothness, by solving a log-degree barrier optimization over the graph Laplacian, instead of computing pairwise connectivity metrics independently.

We validated the method on MEG data and see that it's more robust for short recordings, and produces more consistent individual-level (subject-specific) connectivity estimates. see [paper](https://ieeexplore.ieee.org/abstract/document/11506300). also some other works on motor imagery decoding: [paper](https://www.sciencedirect.com/science/article/pii/S1746809423009709) 

For the implementation, please see this [link](https://mne-rt-org.github.io/mne-rt/modalities.html#sensor-graph), however I think its better to move it to `mne-connectivity` (and modify).  wdyt? @tsbinns @larsoner


drammock:
This looks useful and in-scope.

@payamsash at some point it might be productive to have a conversation (also with @nordme?) about what graph algorithms would be useful to add to MNE-Connectivity, and whether/where it makes sense to start using `networkx` internally for some stuff...  I know it has a reputation of being slow, but I think they've done a lot of work recently on dispatching to different backends so maybe speed is no longer a reason to avoid `networkx` objects.

payamsash:
Just to be clear, Kalofolias is graph construction method same as the existing pairwise estimators so it should land here (`mne_connectivity/graph_learning.py`) if implemented, but `networkx` or `graspologic` are more graph statistics/metrics than per-subject I guess so they actually work one step downstream, when a graph is already constructed.

I think `networkx` could be another PR like "what do we do with the resulting graph?", maybe a lightweight `to_networkx() `(and maybe `to_graspologic()`), which also makes maintenance easier for `mne-connectivity`.

payamsash:
AI disclosure: I used Claude to write `test_graph_learning.py`, then I reviewed it

drammock:
> Just to be clear, Kalofolias is graph construction method same as the existing pairwise estimators so it should land here (`mne_connectivity/graph_learning.py`) if implemented

yeah, we're on the same page about that

> but `networkx` or `graspologic` are more graph statistics/metrics than per-subject I guess so they actually work one step downstream, when a graph is already constructed.
> I think `networkx` could be another PR like "what do we do with the resulting graph?", maybe a lightweight `to_networkx() `(and maybe `to_graspologic()`), which also makes maintenance easier for `mne-connectivity`.

I was also thinking that `.to_networkx()` might be all we want/need. Let's stew on it for a bit, and maybe open an issue to solicit wider input.

drammock:
merging main to hopefully fix a weird CircleCI error

payamsash:
just in case it slipped your attention, thanks @drammock

drammock:
> just in case it slipped your attention, thanks @drammock

thanks for the ping. Plan is for @tsbinns to review this one, but we want to get the 0.9 release out first. One bit of advance feedback is that we'll likely want a short example/tutorial/demo of this method added to the docs.

doc/changes/dev/439.newfeature.rst
```diff
@@ -0,0 +1 @@
+Add :func:`~mne_connectivity.learn_graph` for estimating sparse functional connectivity graphs directly from signal smoothness :footcite:`kalofolias2016learn,shabestari2026shared`, by `Payam S. Shabestari`_.
```
tsbinns:
```suggestion
Add :func:`~mne_connectivity.learn_graph` for estimating sparse functional connectivity graphs directly from signal smoothness, by `Payam S. Shabestari`_.
```
We usually don't include citations in changelogs. Best place to have them is within the function docstring.

doc/references.bib
```diff
@@ -144,6 +144,15 @@ @article{HippEtAl2012
  year = {2012}
 }
 
+@inproceedings{kalofolias2016learn,
```
tsbinns:
```suggestion
@inproceedings{Kalofolias2016,
```

doc/references.bib
```diff
@@ -248,6 +257,17 @@ @book{SekiharaNagarajan2008
  year = {2008}
 }
 
+@article{shabestari2026shared,
```
tsbinns:
```suggestion
@article{ShabestariEtAl2026,
```

doc/api.rst
```diff
@@ -44,6 +44,7 @@ on numpy array inputs.
    :toctree: generated/
 
    envelope_correlation
+   learn_graph
```
tsbinns: Nitpick, but I think `learn_graph` should be at the end, since it's a bit more of a distinct method compared to the others, like `wsmi` is, and these lists aren't alphabetically ordered.

mne_connectivity/graph_learning.py
```diff
@@ -0,0 +1,372 @@
...
+    verbose=None,
+):
+    """Learn a sparse graph from smooth signals.
+
+    Estimates a symmetric, non-negative, zero-diagonal weighted adjacency
+    matrix directly from signal smoothness, rather than from pairwise
+    similarity/coherence computed independently per edge. All edge weights
+    are learned jointly by solving::
+
+        minimize_{w >= 0}  2 * w^T z - alpha * 1^T log(S w) + beta * ||w||^2
```
tsbinns: Would be nice to render this as an equation rather than a code block

mne_connectivity/graph_learning.py
```diff
@@ -0,0 +1,372 @@
...
+    ----------
+    data : array_like, shape ([n_epochs,] n_nodes, n_times) | ~mne.Epochs | generator
+        The data from which to learn a graph. Can be a single 2D array of
+        shape ``(n_nodes, n_times)`` (in which case a single
+        :class:`~mne_connectivity.Connectivity` is returned), a 3D array of
+        shape ``(n_epochs, n_nodes, n_times)``, a :class:`mne.Epochs` object,
+        or a list/generator of 2D arrays or :class:`mne.SourceEstimate` /
+        :class:`mne.VolSourceEstimate` objects (in which case an
+        :class:`~mne_connectivity.EpochConnectivity` is returned, one graph
+        per epoch).
```
tsbinns: We don't need to mention what the output type for each input is here. We tend to specify that in the description of the outputs.

mne_connectivity/graph_learning.py
```diff
@@ -0,0 +1,372 @@
...
+        consistently regardless of the input's physical units or amplitude
+        scale. See Notes.
+    max_iter : int
+        Maximum number of primal-dual iterations. Default ``1000``, which is
+        ample for the default ``alpha=0.5, beta=0.5`` (typically well under
+        100 iterations). Sparser settings (small ``beta``, in particular
+        ``beta=0``) lack a strongly-convex term and converge markedly more
+        slowly, often requiring several thousand iterations; a
+        :class:`RuntimeWarning` is raised if ``tol`` is not reached within
+        ``max_iter`` iterations.
```
tsbinns: We don't usually mention what warning/error a certain choice of parameters might introduce. I think it's enough to say that sparser settings may require more iterations - as you do - and then allow the user to figure out what works for their exact configuration.

mne_connectivity/graph_learning.py
```diff
@@ -0,0 +1,372 @@
...
+    tol : float
+        Relative convergence tolerance on both the edge-weight and
+        node-degree updates. Default ``1e-4``.
+    step_size : float | None
+        Advanced: override the closed-form step size
+        ``1 / (1 + 2 * beta + sqrt(2 * (n_nodes - 1)))``. Convergence is only
+        guaranteed for the default. Default ``None``.
```
tsbinns: What is the case where users would want to play around with this? Is it worth discussing in the notes?

mne_connectivity/tests/test_graph_learning.py
```diff
@@ -0,0 +1,216 @@
...
+    rng = np.random.default_rng(4)
+    X, _ = _make_synthetic_graph_signals(5, 500, rng)
+    conn = learn_graph(X, verbose=False)
+    assert isinstance(conn, Connectivity)
+    assert not isinstance(conn, EpochConnectivity)
+
+
+@pytest.mark.parametrize("n_jobs", [1, 2])
+def test_learn_graph_n_jobs_identical(n_jobs):
+    """Test that n_jobs=1 and n_jobs=2 give identical results."""
```
tsbinns: Parametrising `n_jobs` here is unnecessary. In the `n_jobs=1` case, it's testing that a call to `learn_graph` with the same inputs gives the same results, which is trivial. You could just have an unparametrised test where you compare `n_jobs=1` and `n_jobs=2`.

That would also be enough to check that the output is deterministic, in case that was the idea behind comparing `n_jobs=1` == `n_jobs=1`.

tsbinns (changes requested):
@payamsash Sorry for the delay in getting to this. Mainly minor comments.

I didn't delve too much into the equations, so I don't have any comments there. At least for now nothing stuck out so much it caught my eye.

As @drammock mentioned, some example/tutorial would be nice, even a small thing that just shows the function being used with some data. If you'd like, it can also be a more detailed tutorial about using this in an analysis. It's really up to you. You can have a look at the existing examples for some inspiration.

tsbinns:
Don't worry about the failing test; it's unrelated to this PR and will be fixed upstream.

payamsash:
Thank you @tsbinns, meanwhile I found this unmerged [PR](https://github.com/epfl-lts2/pygsp/pull/57) and would like to tag @kalofolias directly as this is his algorithm. We'd be glad to hear you're happy to see the method added to `mne-connectivity`. Few points:

- Does this implementation looks right to you @kalofolias?
- I think we can add two things from your implementation: 1. automatic alpha/beta selection from a target average node degree 2. sparse (candidate edge restricted) mode for larger graphs (e.g. source-space vertices)
- also as @tsbinns and @drammock mentioned, would be nice to have a example/tutorial on this? do you have a suggestion for that?

payamsash:
@kalofolias Just a kind reminder to have a look at this PR when you get a chance. Thanks!

PR #449, opened 2026-08-12
Fix unit test conda h5py-hdf5 incompatibility and update pip python versions

tsbinns:
Conda test job failing as `h5py` is running against a version of HDF5 which it wasn't built against (see: https://github.com/mne-tools/mne-connectivity/actions/runs/31607822366/job/94151223273).

Wasn't happening yesterday: https://github.com/mne-tools/mne-connectivity/actions/runs/31503504174/job/93819307617

Only difference I can see is the `hdf5` version getting bumped between runs (1.14.6 to 2.2.0).

`environment.yml` doesn't list `netCDF4` as a dependency like `pyproject.toml`, so maybe the initial conda env creation followed by the `pip install -e .` is causing the conflict.

Want to see if adding `netCDF4` as a dependency in `enviropnment.yml` fixes things. Also copying over other version specifiers whilst at it.


tsbinns:
Nope, will have to dig into further :'(

tsbinns:
I just realised I'm not sure why the conda job was running on the older Python version in the first place. Nothing else gets pinned, so it's not like it's an 'old' CI job. We have a Python 3.11 pip job, so I don't see why the conda job shouldn't be bumped to Python 3.13.

tsbinns:
Whilst at it, I added a Python 3.14 job using `mne-main`, and switched the 'old' Python 3.11 run to use `mne-stable`, which seems more fitting.

larsoner:
Sounds good, are you able to adjust the branch protection rule or you need me to do it?

tsbinns:
I don't have perms, please can you update the expected names 🙏🏻

larsoner:
Adjusted reqs and pushed a fix to move to the app version of `circleci-artifacts-redirector-action`, marking for merge-when-green, thanks in advance @tsbinns !

PR #454, opened 2026-08-18
Remove azure and add codecov token

larsoner:
1. Azure doesn't seem to run anymore, and seems redundant with GHA anyway, so let's remove it.
2. `codecov` didn't seem to report on my previous PR, let's see if this fixes it (I doubt it but who knows)


larsoner:
Okay the latest `codecov` report is from 2024. So I think merging this should cause `main` to finally report coverage to codecov, which should then make it report properly on PRs.

@tsbinns feel free to merge if you're happy!

tsbinns (approved):
Didn't even register for me that we were supposed to have azure pipelines and codecov jobs running. Thanks @larsoner!

tsbinns:
👏🏻 
image

Issue #444, opened 2026-08-01
Downstream graph-analysis support


payamsash:
#### Describe the problem
following #439 and the discussion with @drammock: 

`mne-connectivity` currently stops after creating the graph/connectivity matrix and *analysis* (centrality, clustering, community detection, etc.) is left entirely to the user. it's worth deciding whether/how to make it easier to hand results off to some other graph packages.

#### Describe your solution
One easy option could be (`.to_networkx()`, and/or `.to_graspologic()`) as a low-maintenance middle ground.

 Opening this to get wider input. Thanks in advance.

drammock:
cc @ruuskas @nordme @adam2392 happy to hear your thoughts on this

viranovskaya:
I would start with `.to_networkx()` as the smallest useful interoperability layer, with NetworkX remaining an optional dependency imported only when the method is called.

A few parts of the conversion contract seem important:

- Connectivity with epoch, frequency, time or component dimensions should not be silently averaged. The converter should either require an already reduced 2D node-by-node representation or accept an explicit selector/reducer.
- Preserve `names` as node labels and include all nodes, including isolated ones.
- Avoid inferring directedness from method names. An explicit `create_using` or `directed` argument would be safer, with an automatic default only when symmetry is unambiguous.
- For connectivity stored using explicit indices, construct the edge list directly instead of materialising a potentially large dense matrix.
- Store relevant provenance such as the connectivity method and any frequency/time selection or reduction in `graph.graph`.

I would cover directed and undirected connectivity, dense and indexed storage, node labels, isolated nodes, and refusal of unreduced multidimensional data in the initial tests.

If this scope sounds reasonable, I would be happy to prepare a focused first PR for `.to_networkx()` before considering a separate graspologic adapter.

drammock:
let's discuss this further at our meeting on the week of Aug 24th

ruuskas:
I support this idea. Currently, it is up to the user to figure out how to export the connectivity data into NetworkX or similar. The points made by @viranovskaya also sound good to me. 

A possible problem for ease-of-use I see here is that, for example, a `SpectralConnectivity` object contains the frequency dimension, but AFAIK there's currently no direct way to combine the data across frequency bins. The same would be true for the time dimension in `TemporalConnectivity`. If users anyway need to use `.get_data()`, then the utility of e.g. `.to_networkx()` is undermined. 

[...]

Issue #440, opened 2026-07-21
Expected changes for 1.0


tsbinns:
Collating a list of changes that we are expecting to take place in v1.0 release.

## Deprecations started in v0.9
End-users can already prepare for these changes already, based on the work that was started in v0.9, i.e., there is a deprecation period.

- Remove `make_surrogate_data`. `make_surrogate_resting_data` should be used in its place.
- Switch from `n_times` (units of samples) to `duration` (units of seconds) as the way of specifying epoch lengths in `make_signals_in_freq_bands` (`connection_delay` will also be specified in units of seconds rather than samples).

## Anticipated changes
End-users cannot fully prepare for these changes, as they will break MNE-Connectivity ≤ v0.9 code, i.e., there is no deprecation period.

### Separating MNE data object and array-like data inputs (#373)
Separate connectivity-computing functions into those accepting data as arrays versus MNE data objects.

E.g., `vector_auto_regression` currently accepts data as array-likes or `mne.Epochs`. In v1.0, `vector_auto_regression` will only accept `mne.Epochs` as input data, and a new function, `vector_auto_regression_array` would need to be used with array-like data.

The expected affected functions are:
- `envelope_correlation`
- `phase_slope_index`
- `phase_slope_index_time`
- `vector_auto_regression`
- `spectral_connectivity_epochs`
- `spectral_connectivity_time`
- `wsmi`

Furthermore, those functions which accept array-like data as input will return array-like data, not data packaged into a connectivity container class. This is akin to the functions in MNE-Python's `time_frequency` module that accept array-like data (e.g., `mne.time_frequency.psd_array_welch`) returning results as arrays, alongside additional information like the frequency or time bins in the results dimensions.

### Requiring spectral representations of data to be provided (#373)
Functions that rely on a spectral representation of data will no longer accept `mne.Epochs` (or array-like time-series data) as inputs. Instead, the spectral representations that were computed internally in MNE-Connectivity ≤ v0.9 will need to be provided directly by the end-user.

[...]

PR #457, opened 2026-08-25
[ENH] Add .to_networkx() for downstream graph-analysis support

raphbrd:
PR Description
--------------

Add a .to_networkx() method to the BaseConnectivity class for downstream graph-analysis support. The export object is either a networkx.Graph or networkx.DiGraph object. Also, add an example in examples/connectivity_to_graph.py.

Closes issue #444 .


tsbinns:
Discussion/findings with @raphbrd:
- `Graph` with square matrix as input with non-zero tril entries will mirror values to triu when querying edge data, e.g., `get_edge_data(0, 1)` == `get_edge_data(1, 0)` (vice-versa for non-zero triu matrix).
- `Graph` will also mirror tril values to triu when querying edge data, even when triu has non-zero entries (although original data is still stored in the object). Suggest giving a warning if a `Graph` object is requested when data is not a tril/triu matrix.
- `DiGraph` object preserves input values when querying edge data, e.g., `get_edge_data(0, 1)` != `get_edge_data(1, 0)`.
- `DiGraph` object accepts square matrix when tril/triu values are zero. The zero weights for those entries is preserved when querying edge data.
- Suggest having `is_directed=None` as default, allowing whether `Graph` or `DiGraph` object is returned to be determined automatically, e.g.: `Graph` when `con.indices=="symmetric"`; `DiGraph` when `con.indices=="all" or None`; `DiGraph` when `con.indices` is a tuple. Still allow overriding `is_directed` as False/True (flexibility for users to edit the graph object; also with warning about `Graph` being inappropriate for 'full' matrix data).

tsbinns:
Would also be cool once `mne.Info` is stored in connectivity containers (see #440) to add node attributes like channel types, positions with https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.set_node_attributes.html#networkx.classes.function.set_node_attributes

But that can be a subsequent PR

doc/changes/dev/457.newfeature.rst
```diff
@@ -0,0 +1 @@
+Add a :meth:`~mne_connectivity.base.BaseConnectivity.to_networkx` method for downstream graph-analysis support by `Marie-Constance Corsi`_ and `Raphael Bordas`_
```
tsbinns: The `BaseConnectivity` class isn't public API, so I'd suggest the following:

```suggestion
Add a :meth:`~mne_connectivity.base.Connectivity.to_networkx` method to connectivity containers for downstream graph-analysis support, by `Marie-Constance Corsi`_ and `Raphael Bordas`_
```

mne_connectivity/base.py
```diff
@@ -917,6 +918,86 @@ def save(self, fname):
...
+            The exported connectivity data in networkx format. Dimensions are identical
+            to dimensions of the original Connectivity object.
+        """
+        nodelist = self.names
+
+        # check whether the connectivity matrix is directed
+        if is_directed is not None and not isinstance(is_directed, bool):
+            raise TypeError(
+                f"is_directed must be a bool or None, got {type(is_directed)}"
+            )
```
tsbinns: We can use some MNE helpers for this

```suggestion
        _validate_type(is_directed, (bool, None), "is_directed")
```

mne_connectivity/base.py
```diff
@@ -917,6 +918,86 @@ def save(self, fname):
...
+        # or the indices attribute of the connectivity matrix
+        # indices can be: 'all', tuple of array, None, or 'symmetric'
+        method = nx.Graph()
+        if is_directed is None or is_directed:
+            if (
+                self.indices is None
+                or self.indices == "all"
+                or isinstance(self.indices, tuple)
+            ):
+                method = nx.DiGraph()
```
tsbinns: Right now, that nested check could override the user's choice of `is_directed=True` if `indices=="symmetric"`, but I think we should allow the user to have final say. 

```suggestion
        if is_directed is None:
            if self.indices == "symmetric":
                is_directed = False
            else:
                is_directed = True
        if is_directed:
            method = nx.DiGraph()
        else:
            method = nx.Graph()
```

mne_connectivity/base.py
```diff
@@ -917,6 +918,86 @@ def save(self, fname):
...
+        if self.is_epoched:
+            new_shape.append(self.n_epochs)
+        if "components" in self.dims:
+            new_shape.append(len(self.coords["components"]))
+        if "freqs" in self.dims:
+            new_shape.append(len(self.coords["freqs"]))
+        if "times" in self.dims:
+            new_shape.append(len(self.coords["times"]))
+
+        con_matrix = self.get_data(output="dense")  # shape: n_nodes * n_nodes, dims
```
tsbinns, line 1007: Would be good to have a warning that a 'full' matrix is being passed to a `Graph` object instead of a `DiGraph` object. E.g.:

```
if not is_directed:
    if not <all tril or triu entries are zero or NaN> or <data != data.T>:
        warn("Non-symmetric connectivity data is being passed to a `Graph` object. Consider passing `is_directed=True` to use a `DiGraph` object.", UserWarning)
```

Or something like this.

tsbinns (commented):
A couple of comments. This will likely require also updating the tests.

tsbinns:
We should catch if the data is multivariate and raise an error that this behaviour is currently not supported (`get_data("dense")` doesn't work for this kind of data; see #458). This can be addressed in a subsequent PR once #458 is resolved.

mne_connectivity/base.py
raphbrd: It now produces a warning

mne_connectivity/base.py
```diff
@@ -917,6 +918,104 @@ def save(self, fname):
...
+            networkx.Graph, True export date to networkx.DiGraph. None infers the graph
+            type from the connectivity indices (None, 'full', or tuple of arrays yield a
+            directed graph; 'symmetric' yields an undirected graph). Default is None.
+        is_weighted : bool
+            Whether the output graph should be weighted. False corresponds to
+            thresholded connectivity matrices. Default is True.
+
+        Returns
+        -------
+        list of networkx.Graph | networkx.
```
tsbinns: To get this to generate a link (that is doing using a tool called intersphinx), we should add networkx to the intersphinx packages here: https://github.com/mccorsi/mne-connectivity-graph/blob/c7eef0c8ad1816f380c30fc51e5276a2123cd0cb/doc/conf.py#L297

```suggestion
        list of networkx.Graph | networkx.DiGraph
```

mne_connectivity/base.py
```diff
@@ -917,6 +918,104 @@ def save(self, fname):
...
+
+        Multivariate connectivity data cannot be exported to networkx graphs.
+
+        Parameters
+        ----------
+        is_directed : bool | None
+            Whether the connectivity matrix is directed or not. False exports data to
+            networkx.Graph, True export date to networkx.DiGraph. None infers the graph
+            type from the connectivity indices (None, 'full', or tuple of arrays yield a
+            directed graph; 'symmetric' yields an undirected graph). Default is None.
```
tsbinns: We don't usually add info about return types here (since that info is present in the `Returns` section). Also, we can add some nice formatting to the vars.

Line length will need to be fixed.

```suggestion
            Whether the connectivity matrix is directed or not. ``False`` exports data as an undirected graph. ``True`` exports data as a directed graph. ``None`` infers the graph type from the connectivity ``indices``:  ``None``, ``'full'``, or a tuple returns a directed graph; ``'symmetric'`` returns an undirected graph). Default is ``None``.
```

mne_connectivity/base.py
```diff
@@ -917,6 +918,104 @@ def save(self, fname):
...
+
+        # epochs are the first dim: we put them after the two nodes dimension to ensure
+        # proper array reshaping below
+        if self.is_epoched:
+            con_matrix = con_matrix.transpose((1, 2, 0, *range(con_matrix.ndim)[3:]))
+
+        con_matrix_flat = con_matrix.reshape([self.n_nodes, self.n_nodes, -1])
+
+        if not is_directed:
+            if np.any(con_matrix_flat != con_matrix_flat.transpose((1, 0, 2))):
```
tsbinns: We should also consider tril and triu matrices. Right now, these would get a warning even though they are valid undirected graphs.

tsbinns:
## Cases we should test for

### Possible matrix kinds

1. Lower-triangular, where upper-triangular are all zeros or NaNs
2. Upper-triangular, where lower-triangular are all zeros or NaNs
3. Full symmetric, where lower- and upper-triangular are symmetric around the diagonal
4. Full non-symmetric, where lower- and upper-triangular are *not* symmetric around the diagonal

For undirected graphs, cases 1, 2, and 3 are valid. Case 4 is also accepted by networkx, but we should raise a warning.

For directed graphs, cases 1-4 are all valid.

### Tests

#### Main set of tests

The following can all come in one test, with parametrised inputs.

We should test:
- each connectivity container
  - works with each matrix kind
    - for undirected/directed graphs

For these, just give the indices explicitly as tuples of arrays. Don't use `None`, `'all'`, or `'symmetric'` here.

As well as checking the code runs, we should check the edge attributes of some nodes match the original connectivity array data.

We should check a warning is raised for undirected graphs and matrix kind 4.

#### Additional tests

The following can all come in one test, with parametrised inputs.

For the base `Connectivity` container, we should check that the code works with indices as `None`, `'all'`, or `'symmetric'`, and that the edge attributes are as expected based on the input data.

For `None` and `all`, input data should be a full matrix. For `symmetric`, input data should be an upper-triangular matrix with diagonal.

With this, we should also check that `is_directed=None` picks an appropriate graph type.

tsbinns:
Code is looking good! Just want to touch up the example and this should be good to merge (although perhaps with a review from @drammock or @larsoner since I have been involved in the writing).

tsbinns:
Made some edits to the example, mainly removing the example NetworkX analyses. They have a lot of examples/tutorials in their own documentation, so I lean towards more just showing people how to export the data and then directing them towards NetworkX's docs.

In a subsequent PR if we add attrs to the graphs like time/freq bin labels, we could add something on this to show more how the graph objects map to the original connectivity containers.

mne_connectivity/tests/test_connectivity.py
```diff
@@ -470,3 +484,201 @@ def test_get_data_complex(indices):
...
+    # Case 2: Upper-triangular, where lower-triangular are all zeros or NaNs
+    indices_u = np.triu_indices(n_nodes)
+    # Testing both cases: export to either directed or not directed graphs
+    for indices in [indices_l, indices_u]:
+        conn = conn_cls(
+            data=correct_numpy_input, n_nodes=n_nodes, indices=indices, **extra_kwargs
+        )
+        G = conn.to_networkx(directed=directed)
+
+        expected_shape = correct_numpy_shape.copy()
```
drammock:
```suggestion
        # networkx will return an *array* of graph objects
        # (one per epoch/freqbin/timepoint/etc), with shape:
        expected_shape = correct_numpy_shape.copy()
```

mne_connectivity/tests/test_connectivity.py
```diff
@@ -470,3 +484,201 @@ def test_get_data_complex(indices):
...
+        expected_shape.pop(node_axis)
+
+        if len(expected_shape) == 0:
+            assert isinstance(G, nx.DiGraph if directed else nx.Graph)
+        else:
+            assert_array_equal(G.shape, expected_shape)
+            flat_G = G.flatten()
+            assert isinstance(flat_G[0], nx.DiGraph if directed else nx.Graph)
+            assert_array_equal(len(flat_G[0].edges.data("weight")), expected_n_weights)
```
drammock: better to test the actual *values* of the weights against the input data, not just the length

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+        """Export the connectivity data to networkx format.
+
+        Parameters
+        ----------
+        directed : bool | None
+            Whether the connectivity matrix is directed or not. ``False`` exports data
+            as an undirected graph. ``True`` exports data as a directed graph. ``None``
+            infers the graph type from the connectivity ``indices``:  ``None``,
+            ``'all'``, or a tuple returns a directed graph; ``'symmetric'`` returns an
+            undirected graph. Default is ``None``.
```
drammock:
```suggestion
            infers the graph type from the connectivity object's ``indices`` attribute:
            if ``indices`` is ``None``, ``'all'``, or a tuple, this returns a directed
            graph; if ``indices`` is ``'symmetric'`` this returns an undirected graph.
            Default is ``None``.
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+
+        Returns
+        -------
+        graphs : networkx.Graph or networkx.DiGraph | array of networkx.Graph or networkx.DiGraph
```
drammock:
```suggestion
        graphs : networkx.Graph | networkx.DiGraph | array of networkx.Graph | array of networkx.DiGraph
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+            The connectivity data in networkx format. If additional dimensions (e.g.,
+            frequencies, times, etc...) are present, a graph is returned for each of
```
drammock:
```suggestion
            frequencies, times, etc.) are present, a graph is returned for each of
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+        # check whether the connectivity matrix is directed
+        _validate_type(directed, (bool, None), "directed")
+
+        # infer directionality of the graph from either the is_directed arg
+        # or the indices attribute of the connectivity matrix
+        if directed is None:
+            if self.indices == "symmetric":
+                directed = False
+            else:
+                directed = True
```
drammock:
```suggestion
            directed = self.indices != "symmetric"
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+        if directed:
+            container = nx.DiGraph
+        else:
+            container = nx.Graph
```
drammock:
```suggestion
        container = nx.DiGraph if directed else nx.Graph
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+        # check the dimensions of connect_matrix
+        new_shape = []
+        if self.is_epoched:
+            new_shape.append(self.n_epochs)
+        if "components" in self.dims:
+            new_shape.append(len(self.coords["components"]))
+        if "freqs" in self.dims:
+            new_shape.append(len(self.coords["freqs"]))
+        if "times" in self.dims:
+            new_shape.append(len(self.coords["times"]))
```
drammock: it's awkward that we need to do this to get the expected shape of the data. We should add a helper / attribute to get this directly from the object in one line.

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+
+        con_matrix = self.get_data(output="dense")  # shape: n_nodes * n_nodes, dims
```
drammock:
```suggestion
        con_matrix = self.get_data(output="dense")  # shape: ([epochs,] n_nodes, n_nodes, [...dims])
```

mne_connectivity/base.py
```diff
@@ -956,6 +957,104 @@ def save(self, fname):
...
+        if not new_shape:  # no extra dims
+            con_matrix = con_matrix[..., np.newaxis]  # add a temporary extra dim
+
+        # epochs are the first dim: we put them after the two nodes dimension to ensure
+        # proper array reshaping below
+        if self.is_epoched:
+            con_matrix = con_matrix.transpose((1, 2, 0, *range(con_matrix.ndim)[3:]))
+
+        con_matrix_flat = con_matrix.reshape([self.n_nodes, self.n_nodes, -1])
```
drammock: it's worth considering if we can avoid the transposing and reshaping here, and instead iterating over the correct axes of `con_matrix` to get the 2D slices that `networkx.from_numpy_array` wants.

I think that involves:
- promoting `con_matrix` to have singleton dimensions where needed, so it always has shape `(epos, chs, chs, comps, freqs, times)`
- creating a corresponding output container array (with `chs, chs` replaced by one dim for holding the graph)
- constructing an iterator/generator of indices over the non-`chs` axes
- looping over that iterator/generator to extract the adjacency matrix to pass to nx
- using those indices to put the resulting graph object in the corresponding cell of the output array

It may not be fewer lines of code but it feels less error-prone to me. I can take a swing at it if it sounds daunting to you, just LMK

PR #459, opened 2026-08-26
Allow multivariate data to be returned in a dense form

### LINKED ISSUE #458: Allow multivariate data to be returned from `get_data()` in a 'dense' form
Currently, multivariate data can only be returned in a raveled form as if we consider each row/column in the connectivity matrix as a single channel, a single set of multivariate results would correspond to multiple channels (entries in the matrix), and there could also be cases where multiple results correspond to the same channels (entries in the matrix).

This could be changed so that the matrix is considered in terms of nodes (i.e., the set of channels that make up the multivariate seed/target) rather than the single channels.

I have made some progress on the logic for this as part of the visualisation improvements, and now it would also be relevant to #457 for allowing multivariate data to be converted to a `networkx` graph object.

tsbinns:
Fixes #458

Remaps multivariate indices into the unique nodes of the connectivity, allowing multivariate connectivity to be represented in a dense form.

### DIFF
--- a/examples/handling_ragged_arrays.py
+++ b/examples/handling_ragged_arrays.py
@@ -86,8 +86,80 @@ from mne_connectivity import spectral_connectivity_epochs
 # MNE-Connectivity combine information across the different channels into a
 # single (time-)frequency-resolved connectivity spectrum, regardless of the
 # number of seed and target channels, so ragged arrays are not a concern here.
+
+########################################################################################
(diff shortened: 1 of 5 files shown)


larsoner (approved):
Works for me, marking for merge-when-green (and restarting the flaky CI). Thanks in advance @tsbinns !

Issue #335, opened 2025-09-18
Add a `picks` parameter to `vector_auto_regression` and `envelope_correlation`


tsbinns:
#### Describe the problem
`vector_auto_regression` and `envelope_correlation` will compute results for all channels, meaning users must first pick channels before passing this to the functions. This goes against the API of similar functions in MNE where a `picks` parameter is present.

Also since #334, bad channels will always be excluded from computations in these functions, but many other functions in MNE allow the `picks` parameter to have users explicitly include bads.

#### Describe your solution
Add a `picks` paramater to these functions, with the same functionality as elsewhere in MNE.

I'd say `picks` is more appropriate here than `indices` (which is used for a lot of functions in MNE-Connectivity), since signal envelope is a within-channel measure, and all-to-all connectivity is implied for VAR models.

PR #461, opened 2026-08-28
Update indexing to make consistent

tsbinns:
Fixes #372

Possible indices kinds will be:
- tuples of arrays for custom seed-target combinations
- `'all'` for all-to-all connectivity (as current); `None` alias will be removed
- `'upper'` and `'lower'` for upper- and lower-triangular parts (excluding diagonal); replaces current `'symmetric`

For connectivity-computing functions accepting the `indices` parameter, valid inputs are:
- tuples of array-likes, always.
- `'all'`, always.
- `'lower'` and `'upper'`, when missing values for a dense connectivity matrix can be inferred based on the existing values (currently nowhere this isn't true).

For multivariate methods, indices will always need to be specified as tuples.

In connectivity containers, where `get_data()` is called with `'dense'` output and `indices` is  `'upper'` or `'lower'`, we will attempt to fill the missing values. If this is not possible (`method` is unknown, or it is known and the transformation is not possible), behaviour is determined by the `missing` parameter. Default is to raise an error, but can be set to a float to fill missing values without trying to transform anything.
When `indices` is `'all'`, no transformation is needed (just reshaping).
When `'indices'` is a tuple, no transformation is needed if the indices correspond to all-to-all connectivity. Otherwise, `missing` should be specified as a float (i.e., won't try any fancy logic to know what should be filled in, at least for now).

### DIFF
--- a/mne_connectivity/base.py
+++ b/mne_connectivity/base.py
@@ -455,11 +455,9 @@ class BaseConnectivity(EpochMixin):
         metadata=None,
         **kwargs,
     ):
-        if isinstance(indices, str) and indices not in ["all", "symmetric"]:
-            raise ValueError(
-                'Indices can only be "all", "symmetric", or a list of tuples. '
-                f"It cannot be {indices}."
-            )
+        _validate_type(indices, (str, tuple), "indices")
+        if isinstance(indices, str):
+            _check_option("indices", indices, ["all", "lower", "upper"])
 
         # prepare metadata pandas dataframe and ensure metadata is a Pandas
         # DataFrame object
@@ -594,10 +592,6 @@ class BaseConnectivity(EpochMixin):
 
         # get the number of estimated nodes
         self._get_num_connections(data)
-        if self.is_epoched:
-            data_len = data.shape[1]
-        else:
-            data_len = data.shape[0]
 
         if isinstance(indices, tuple):
             # check that the indices passed in are of the same length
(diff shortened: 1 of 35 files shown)


tsbinns:
My idea for the transformation tests would be to compute connectivity with `indices` as `'lower'` and `'all'`, and compare dense outputs are equivalent. Not currently possible for all methods (e.g., envelope correlation does not have `indices` param to control this; see #335), but could check with what is possible currently.

larsoner:
Sounds good @tsbinns ! In the meantime CIs are red as well (but you probably noticed 😄 )

tsbinns:
> Sounds good @tsbinns ! In the meantime CIs are red as well (but you probably noticed 😄 )

Yeah, haven't touched any of the existing tests yet

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -1575,110 +1564,42 @@ def test_spectral_connectivity_time_freqs(method, freqs, mode):
...
         con_matrix = con_matrix.mean(axis=0)
 
-    # signals are perfectly phase-locked, connectivity matrix should be
-    # a lower triangular matrix of ones
-    assert np.allclose(con_matrix, np.tril(np.ones(con_matrix.shape), k=-1), atol=0.01)
-
-
-@pytest.mark.parametrize("method", ["coh", "imcoh", "cohy", "plv", "pli", "wpli"])
-@pytest.mark.parametrize("mode", ["cwt_morlet", "multitaper"])
-def test_spectral_connectivity_time_resolved(method, mode):
```
tsbinns, line 1591: Removed this test, because it's essentially a copy of the following test (which had to be updated).

mne_connectivity/spectral/tests/test_spectral.py
```diff
@@ -1575,110 +1564,42 @@ def test_spectral_connectivity_time_freqs(method, freqs, mode):
...
-        )
+    # signals are perfectly phase-locked, connectivity matrix should be all ones
+    assert np.allclose(con_matrix, 1.0, atol=0.01)
 
 
 @pytest.mark.parametrize("method", ["coh", "imcoh", "cohy", "plv", "pli", "wpli"])
 @pytest.mark.parametrize("mode", ["cwt_morlet", "multitaper"])
 @pytest.mark.parametrize("padding", [0, 1, 5])
 def test_spectral_connectivity_time_padding(method, mode, padding):
     """Test time-resolved spectral connectivity with padding."""
```
tsbinns, line 1581: Had to tweak this test. It was testing connectivity was higher between expected signals than to noise signals (in upper-triangular parts). However, the upper-triangular parts were previously just 0. With the new indexing, that is no longer the case, and this test was failing (specifically for methods that discarded zero time lag info).

Have fixed it by adding some delay to the simulated signals, and make sure the indices we are comparing are valid.

mne_connectivity/base.py
```diff
@@ -468,6 +472,19 @@ def __init__(
...
+                bad_indices = f"'{indices}'"  # must be a str
+            elif not _check_if_multivariate_indices(indices):
+                bad_indices = "a tuple of non-nested arrays"  # must be non-nested
+            if bad_indices:
+                raise ValueError(
+                    "`components` are present in `kwargs`, which is a term reserved "
+                    "for multivariate connectivity methods. However, `indices` does "
+                    "not match the format for multivariate methods. Expected a tuple "
+                    f"of nested arrays, got {bad_indices}."
+                )
```
tsbinns, line 487: Reminder to add test coverage for this

tsbinns:
#467 will clean up some of the plotting diff that isn't tied directly to the indices changes.

mne_connectivity/base.py
tsbinns: Done

tsbinns:
@larsoner @drammock This is ready for review. Have updated the initial comment so that it's up to date with the key changes. Have also documented the knock-on effects in the changelog entries.

It is a big diff, but it does affect all parts of the package really. A decent amount of the diff is also extending the plotting tests to make sure the indices changes play nicely there.

larsoner (approved):
I did not look super thoroughly but I did look at the test changes and think they make sense. I think we can test those plus the examples to make sure things are working okay. Good to go from my end. @drammock if you want to look feel free, otherwise @tsbinns I would say feel free to merge in a couple of days (unless Dan asks for more time)

mne_connectivity/tests/test_connectivity.py
```diff
@@ -250,25 +282,218 @@ def test_get_multivariate_data():
...
+        with pytest.raises(
+            ValueError,
+            match="Cannot fill missing values for connectivity data for the method",
+        ):
+            con_bad_meth.get_data()
+
+
+@pytest.mark.parametrize("indices", ["lower", "upper"])
+def test_make_unknown_method_full(indices):
+    """Test that filling missing values in unknown methods errors."""
```
tsbinns: Incorporated this test into one above, but forgot to remove.

tsbinns:
Thanks for checking @larsoner! Sure, will leave open for a few days.

PR #460, opened 2026-08-28
Extend plotting features

tsbinns:
Partially closes #266.
Addresses point 2 of that issue, and partly point 1 for the new features. Will add plotting of multivariate components for the existing plotting functionality in a separate PR. Together that would close #266.

Currently missing unit tests for the new features.

PR #467, opened 2026-09-16
Update plotting

tsbinns:
Updates to recent plotting changes. Will clean up some of the #461 diff.

Main changes:
- Don't return axes from plotting functions. This wasn't documented, and spectrum/TFR plotting in MNE doesn't return axes.
- For matrix plots, show components on different figures.
- Have image plotting functionality directly in `plot_spectrotemporal_connectivity` (was nested in `_plot_image_connectivity`) that only the spectro-temporal plotting called.
- Fix to selection logic when `picks` not None.

Also docstring updates to go along with this.


tsbinns:
@drammock @larsoner Gonna self-merge this when green, as it just cleans up some things from #460.

PR #472, opened 2026-09-19
Add SPEC-0 dependency bumping

tsbinns:
Adds SPEC-0-like dependency version bumping to some dependencies (see https://github.com/mne-tools/mne-tools/pull/52).

One question is how MNE-Python gets handled. Do we also go with the 'what was available 2 years ago' logic, or do we treat it like MNE-Python handles Python vers, and support for >= 3 year period? So far, MNE-Python is not being auto-bumped by the action.

Also, we don't have the towncrier autofix to add PR numbers (see https://github.com/mne-tools/mne-tools/issues/53), or the syncing of dependencies to `environment.yml` (see https://github.com/mne-tools/mne-tools/issues/50 for a pending question about this), so currently there is some manual process required.

Will leave as draft until this is discussed.


tsbinns:
Also bumps min Python to 3.11, which is overdue. The knock-on effect is minimal, so I stuck it in here rather than a separate PR.

tsbinns:
For the dependencies chosen now for auto-bumping, I'm just following what is in MNE-Python.

We don't have a proper 'old' CI job here, so there may be some other packages we find need to be auto-bumped, but until we have that implemented, this should be fine.

tsbinns:
Have also switched to use double ticks consistently in `pyproject.toml`, because the action was changing this from single ticks to double ticks, makign it hard to see the proper diff from the action.

tsbinns:
Would-be diff looks good IMO: https://github.com/mne-tools/mne-connectivity/actions/runs/35462423119/job/105948542916?pr=472

tsbinns:
Will keep as draft until upstream PR https://github.com/mne-tools/mne-tools/pull/52 is merged and proper action path can be set.

https://github.com/mne-tools/mne-connectivity/pull/473 would also avoid need to manually rename towncrier files.

tsbinns:
https://github.com/mne-tools/mne-tools/issues/50 has been sorted, so we can introduce automatic dep syncing to `environment.yml` (will open a separate PR).

tsbinns:
> One question is how MNE-Python gets handled. Do we also go with the 'what was available 2 years ago' logic, or do we treat it like MNE-Python handles Python vers, and support for >= 3 year period? So far, MNE-Python is not being auto-bumped by the action.

Discussed w/ @larsoner. Conclusion was to not auto-bump and potentially support >2 year period if not too much burden, but bump after 2 years if we can clean up ugliness (e.g., would suggest bumping from 1.6 → 1.8 (released Aug 2024) which would allow us to clean up some docstring warnings about complex coeffs in spectrum/TFR objects).

